Issue #42, opened 2019-01-30
Setup a staging site


muffinresearch:
Since we need to iron out some CSP config issues would it be possible to deploy in the same way that we'd deploy to production on a staging site.

Maybe extensionsworkshop.allizom.org could be used or something else? This should be built from master. This way we can iron out the CSP config issues and any other deployment wrinkles ahead of the site being live on https://extensionsworkshop.com.

muffinresearch:
Once this is setup we can disable the gh-pages builds because they won't work with the assets pipeline anyway.

caitmuenster:
Ok, it looks like the stage environment is set up! 🎉 @bqbn, if you could get the basic site up on stage by EOD Monday, that would be great. @muffinresearch will be back from PTO on Tuesday and can start fine-tuning the CSP when he gets back.

bqbn:
The site is up: https://extensionworkshop.allizom.org/

"master" branch is set up to push to the above site automatically.

PR #44, opened 2019-01-31
Push master branch to -stage env

bqbn:
Per https://github.com/mozilla/extension-workshop/issues/42


muffinresearch:
The deploy script will now need a `yarn install` before the jekyll build is run. I'm changing the asset builds in #45 to avoid the need to also install/run gulp to compile the js etc so you might want to wait for that to land.

Issue #15, opened 2018-11-29
Change link location for "Get Started" buttons


MeridelW:
These"Get Started" buttons currently go to the AMO home page. 
![screen shot 2018-11-29 at 2 16 05 pm](https://user-images.githubusercontent.com/33848177/49249361-8b983780-f3e1-11e8-831e-a7676ca277b0.png)

They should take the user to the "Register or Log in" link, instead:
: https://accounts.firefox.com/oauth/signin?scope=profile&state=86afa67a0b6b78ddcfe98977e2950b074976375c450eea199a82984775481dce%3AL2VuLVVTL2ZpcmVmb3gvI2xvZ2lu&redirect_url=https%3A%2F%2Faddons.mozilla.org%2Fapi%2Fv3%2Faccounts%2Fauthenticate%2F&client_id=3a1f53aabe17ba32

muffinresearch:
@caitmuenster I'm not sure how to action this because these links are generated so we can't link to the link in the issue because it won't actually work.

Maybe this is something that can be tied up when the work to sort out the logged-in/out changes happen over on add-ons server, e.g. if linking to the devhub directly automated a login that would probably be what we need for this too.

caitmuenster:
@muffinresearch Oh, ok! Let me remove the blocker then -- I thought this was just a matter of updating the URL. 

I'm ok with sending folks over to AMO to log in.

muffinresearch:
I think we can solve this by linking directly to https://addons.mozilla.org/en-US/developers/addons as this will start the login process.

caitmuenster:
WFM!

PR #65, opened 2019-02-06
Updated "log in/sign up" link

caitmuenster:
Now links to https://addons.mozilla.org/developers/addons

Ref #15


caitmuenster:
Oops -- just realized this does not actually address the issue. Another PR is coming.

muffinresearch:
> Oops -- just realized this does not actually address the issue. Another PR is coming.

I think this is ok isn't it? If you don't have an account you can click the "create account" link.

muffinresearch:
> > Oops -- just realized this does not actually address the issue. Another PR is coming.
> 
> I think this is ok isn't it? If you don't have an account you can click the "create account" link.

Ah I see there's other links. NVM

PR #91, opened 2019-02-13
Enable pushing to -prod

bqbn:
With this change, tagging a release will trigger pushing to -prod.


caitmuenster:
Hey @muffinresearch, would you have a moment to review this today or tomorrow?

bqbn:
It causes a push to -prod on the next tag (and the following tags as well).

PR #105, opened 2019-02-21
Setup GA

### LINKED ISSUE #57: Configure Google Analytics
Once we have a site id to use, it needs to be setup. At the same time we should use a hash for the inline-js in the CSP so that we don't need `unsafe-inline` for `script-src`.

muffinresearch:
Fixes #57

This should enable GA on stage and prod. We'll need to double check the hash is working correctly once this is deployed on stage.


muffinresearch:
I'm going to land this so I can do manual testing on stage.

PR #112, opened 2019-02-22
Add focus styling for a11y

### LINKED ISSUE #110: There are no focus styles
Navigating with a keyboard is showing no focus styles. This is a big hit to accessibility.

To fix this we need to re-instate default focus styles, or alternatives which have likely been turned off.

muffinresearch:
Fixes #110 

This add focus styling to everything that can be tabbed to. The main button focus styling is the exact same styling used on AMO. The link styling is just making use of text-decoration. 

Some cases I defaulted to a default outline to minimize the effect e.g. on logos and the mozilla logo, and social icons.

### DIFF
--- a/_assets/css/_main.scss
+++ b/_assets/css/_main.scss
@@ -73,9 +73,43 @@ small {
   min-width: rem-calc(132);
 }
 
-a:focus,
-button:focus {
+.site-footer .mzp-c-footer-legal ul > li > a:focus,
+a:focus {
   outline: none;
+  text-decoration: underline;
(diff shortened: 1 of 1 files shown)


_assets/css/_main.scss
```diff
@@ -762,7 +796,7 @@ html {
     a {
       display: inline-block;
       width: 100px;
-      height: 32px;
+      height: 28px;
```
muffinresearch: This changes was to have the height of the link match the image.

PR #130, opened 2019-03-07
Edits to content guidelines 

andrewtruongmoz:
@caitmuenster

### DIFF
--- a/content-guidelines/extension-subtitle.md
+++ b/content-guidelines/extension-subtitle.md
@@ -42,24 +42,19 @@ To determine your keywords, think about what words a person would use when searc
 <div class="example" markdown="1">
 <header><h5>Examples:</h5></header>
 
-**Name:** Facebook Container
-
+**Name:** Facebook Container  
 **Subtitle:** Prevent Facebook from tracking you around the web.
 
-**Name:** Remembear
-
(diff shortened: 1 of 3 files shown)


muffinresearch:
> This looks good, but I'm not sure why the Circle CI build is failing.
> 
> @muffinresearch, could you take a quick look?

Prettier needs to be run after making edits to the markdown. This ensures a consistent format. You can do this by running `yarn prettier`, but this requires node + yarn and friends to be installed. I'll see about adding something to do this as a follow-up commit so that we don't need to rely on people to run a command just to do some edits to markdown.

MeridelW:
@muffinresearch Can you please share screenshots of the latest pages so I can do final quality check? Thank you!

muffinresearch:
@caitmuenster if the tests are failing those need to be fixed before branches land. Otherwise as is the changes won't be deployed. 

Mainly we seem to have issues with running prettier, so I'd like to understand if anyone is editing the content can run the command prior to committing and if not we probably need to re-think what we're doing there.

muffinresearch:
I've manually fixed prettier in https://github.com/mozilla/extension-workshop/commit/52006515f8eb904cc1182cb8dee6980ba2832e05 so this can be deployed.

Issue #131, opened 2019-03-26
Add dev environment 


caitmuenster:
As part of Phase II (which kicked off today), we'd like to add a dev environment so we can review and QA updates before pushing them to staging and production. 

There's no rush but if we could get this set up by April 15, that would be great. 

cc: @bqbn, @muffinresearch

bqbn:
@caitmuenster could you clarify the difference between -dev and -stage environments for this project?

I thought that we set up -stage so that we could review QA updates before they're pushed to -prod. Why in addition to -stage, do we need a -dev environment?

caitmuenster:
Good question. My understanding is -stage is there to check over before pushing to -prod, and -dev would be to review/evaluate before we even get to -stage. 

@lancecummings, am I capturing that correctly? Would love to hear your thoughts.

lancecummings:
@caitmuenster we could definitely use stage for this. I don't think the dev request came from us originally, but from my memory, there was the suggestion that you might need the stage environment to continue reviewing and publishing content on the existing architecture, while we simultaneously work on testing the new architecture on the dev environment.

muffinresearch:
One plus I can see about having a hosted -dev environment it could be used to show pages that aren't published on stage or production. This would make it easier to check content as right now we're reliant on someone who can run the site to take screenshots for anyone who doesn't have the development enviroment running locally.

bqbn:
>  This would make it easier to check content as right now we're reliant on someone who can run the site to take screenshots for anyone who doesn't have the development environment running locally.

I'm trying to understand, why can't the changes be pushed to -stage env so that all team members can see? Is it something in the current workflow that prevent us from doing that?

muffinresearch:
> > This would make it easier to check content as right now we're reliant on someone who can run the site to take screenshots for anyone who doesn't have the development environment running locally.
> 
> I'm trying to understand, why can't the changes be pushed to -stage env so that all team members can see? Is it something in the current workflow that prevent us from doing that?

If we do that, stage won't then be representative of what will be released into production. Which is not ideal.

caitmuenster:
@muffinresearch If we take the content guidelines out of consideration, would you be comfortable with Lance testing new code on -stage? If not, let's set up -dev; I don't think the work we're doing on the content guidelines alone justifies setting up a new environment.

muffinresearch:
[...]

PR #166, opened 2019-04-15
Push to dev

bqbn:
In addition to pushing "master" to -stage environment, push it to -dev as well. Non master, i.e. tagged release is pushed to -prod as usual.


bqbn:
Required to fix #131.

https://github.com/mozilla-services/cloudops-deployment/pull/3080 is also needed to create the infrastructure for the -dev environment.

caitmuenster:
@muffinresearch could you review this patch?

.circleci/config.yml
```diff
@@ -65,16 +65,38 @@ jobs:
           command: |
             # show env variables but not the AWS secrets!
             env | egrep -v AWS_
```
muffinresearch: I know this wasn't part of this patch, but it would probably be a good idea to remove this line in case a sensitive env var that doesn't start with `AWS_` is added in the future.

bqbn: The line was there for troubleshooting purpose. I'm OK with removing it.

PR #186, opened 2019-06-13
Page templates

lancecummings:
@muffinresearch Here's a new pull request to deal with just the presentation layer of the search along with new template updates including the styling and functionality of the sidebar navigation.


_includes/nav.html
```diff
@@ -0,0 +1,95 @@
...
+                
+                {% for cat in item.categories %}
+                    <div class="category">
+                        <h6>{{ cat.category }}</h6>
+                        {% if cat.pages %}
+                            <ul>
+                                {% for pg in cat.pages %}
+                                    {% if pg.subpageitems %}
+                                        <li class="{% if page.url == pg.url %}is-active{% endif %}">
+                                            <a href="{{ pg.url | relative_url }}"{% if pg.target %} target="{{ pg.target }}"{% endif %}>{{ pg.title }}</a>
```
muffinresearch: If target could be `_blank` and it's an external URL this must use `rel="noopener referrer"` to avoid leaking the `window.opener` object for security. See https://mathiasbynens.github.io/rel-noopener/ for more information.

This kind of thing could really use a unit test so that we check the output, it would be easy to overlook when content gets updated.

_includes/nav.html
```diff
@@ -0,0 +1,95 @@
...
+                                            <ul class="section">
+                                                {% for section in pg.subpageitems %}
+                                                    <li><a href="#{{ section.id }}">{{ section.title }}</a></li>
+                                                {% endfor %}
+                                            </ul>
+                                        </li>
+                                    {% else %}
+                                        <li class="{% if page.url == pg.url %}is-active{% endif %}"><a href="{{ pg.url | relative_url }}"{% if pg.target %} target="{{ pg.target }}"{% endif %}>{{ pg.title }}</a></li>
```
muffinresearch: As above re `target="_blank"`

documentation/develop.md
```diff
@@ -1,8 +1,38 @@
...
+<section class="panel-collapse module">
+<article class="module-content grid-x grid-padding-x image-with-caption">
+<div class="cell small-6" markdown="1">
+
+![Screenshot Remembear]({% asset "content-guidelines/example-remembear.png" @path @optim %} "Screenshot Remembear") Half width image (above)
+
+</div>
+<div class="cell small-6" markdown="1">
+
+![Screenshot Remembear]({% asset "content-guidelines/example-remembear.png" @path @optime %} "Screenshot Remembear") Half width image (above)
```
muffinresearch: typo `@optime`

_includes/nav.html
lancecummings: I've added the noopener referrer attribute to the template, so it doesn't need to be remembered when updating content.

Issue #191, opened 2019-06-21
Dev environment is not using the unpublished flag.


muffinresearch:
The development environment should be showing the unpublished content whereas stage should not. As it stands neither env is showing the unpublished content. 

To do that the build needs to be passed an `--unpublished` flag only for the **dev** build.

See https://jekyllrb.com/docs/configuration/options/#build-command-options

muffinresearch:
Right now the stage and dev deploys are all being pushed up from the same build, so this needs updating.

bqbn:
Fixed by https://github.com/mozilla/extension-workshop/pull/201.

PR #201, opened 2019-06-26
Build unpublished contents and publish them to -dev env

bqbn:
Per https://github.com/mozilla/extension-workshop/issues/191

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,22 +157,14 @@ workflows:
   version: 2
   build-test-and-deploy:
     jobs:
-      - build-unpublished:
-          filters:
-            branches:
-              only: master
+      - build-unpublished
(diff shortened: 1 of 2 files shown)


muffinresearch:
Hmm looks like the latest run in circle didn't complete.

bqbn:
It didn't start because it is configured to build only "master" and tags. Branches are not built.

I'll merge and fix issues if there are any.

bqbn:
I made a commit so that it builds for all branches, and deploy only for "master" and tags. I think this helps catch errors (if any) when they're still in branches.

bqbn:
Closes https://github.com/mozilla/extension-workshop/issues/191

muffinresearch:
Thanks @bqbn this is working well 👍

PR #202, opened 2019-06-27
Fix #192

### LINKED ISSUE #192: Allow for deploying a specific tag to stage without deploying that tag to prod.
In case we need to bugfix when work is in progress, it would be helpful to add the facility for a specific tag to be pushed to stage in-case we need to stage a specific tag ahead of pushing it to production.

bqbn:
Allow for deploying a specific tag to stage.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,8 +112,6 @@ jobs:
       - run:
           name: extension workshop deployment (unpublished)
           command: |
-            # push master branch to -dev env per
-            # https://github.com/mozilla/extension-workshop/issues/131
             AWS_ACCESS_KEY_ID="${DEV_AWS_ACCESS_KEY_ID}"                 \
             AWS_SECRET_ACCESS_KEY="${DEV_AWS_SECRET_ACCESS_KEY}"         \
             DISTRIBUTION_ID="${DEV_DISTRIBUTION_ID}"                     \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -169,7 +172,6 @@ workflows:
           requires:
             - build
           filters:
-            branches:
-              only: master
             tags:
-              only: /^v?[0-9]+(\.[0-9]+)*$/
+              # match tags such as "v1.1.1" or "anything-stage"
+              only: /^v?[0-9]+(\.[0-9]+)*$|(?:-stage)$/
```
muffinresearch: Based on the circle docs I'm not sure this will work. See https://circleci.com/docs/2.0/workflows/#using-regular-expressions-to-filter-tags-and-branches.

If the tag was `foo-stage` this regex only matches the suffix. Since their setup needs an exact match, you'll need to extend the regex to match anything ahead of the stage suffix. See the example given in their docs:

> For example, only: /^config-test/ only matches the config-test tag. To match all tags starting with config-test, use only: /^config-test.*/ instead.

I think to fix this we could also keep the numeric prefix so the regex could be:

`^v?[0-9]+(\.[0-9]+)*(?:-stage)?$`

This should then completely match all of the following:

- v1.2
- 1.2
- v1.2-stage
- 1.2-stage

bqbn: Hmm, this is strange, because it matched this branch, i.e. `tag-pushing-stage` fine, and ran the deployment step to push to -stage, as can be seen https://circleci.com/gh/mozilla/extension-workshop/537?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I guess we can do what you proposed to be sure.

muffinresearch: That seems wrong doesn't it? I wouldn't expect a deploy to happen for a PR branch? Is that running because there's now no branch filter?

bqbn: Heh, I guess this branch name inadvertently showed that it didn't take an exact match for the tag filter to work. However, I think you're right that branches should not trigger deployment no matter what their names are.

I've updated the PR with comments. Please review again.

PR #207, opened 2019-07-09
Page builds

lancecummings:
@muffinresearch - I've built out a bunch of pages for the site, and in the process streamlined a couple things. Namely: 
- automated the table of contents creation on pages based on the pages.yaml data
- automated the creation of 'next steps' links based on the pages.yaml data


muffinresearch:
A gigantic 5k line patch is going to take some time to review. Please could you consider breaking up the changes further in future PRs?

As a yardstick a 500 line patch is around the max to aim for per branch, it's just a guide and sometimes that'll be exceeded, but more frequently posted smaller patches are a lot easier to review than one gigantic one that changes lots of different things.

community.md
```diff
@@ -37,27 +32,55 @@ Short and brief description of the sort of content you will find on this page.
...
+
+</div>
+</article>
+</section>
+
+<!-- END: Single Column Body Module -->
+
+<!-- Single Column Body Module -->
+
+<section id="connect-with-the-community" class="module">
```
muffinresearch: The `section` elements are a missing a heading here. A section should have a heading as direct child of section.

See https://www.w3.org/TR/html52/sections.html#the-section-element

> The section element represents a generic section of a document or application. A section, in this context, is a thematic grouping of content. Each section should be identified, typically by including a heading (h1-h6 element) as a child of the section element.

muffinresearch:
Some of the images looks like illustrations so they might be better quality if svg was used rather than jpg:

- _assets/img/documentation/develop/EnterpriseExtensions_promo.jpg
- _assets/img/documentation/develop/MultipleDevicesAndBrowsers_fullwidth.jpg
- _assets/img/documentation/develop/PowerUpExtensions_fullwidth.jpg

Some of the images are rather large - please check and see if the optimization steps in the build process are going to be enough. There's a video background image that's 1920px wide, that probably doesn't need to be that big, and since the current image optimizations at build won't change the dimensions you'll need to check that the starting image is appropriately sized.

_data/pages.yaml
```diff
@@ -27,51 +61,203 @@
...
+                - title: "Avoid unnecessary permissions"
+                  id: "avoid-unnecessary-permissions"
+                - title: "Request permissions at runtime"
+                  id: "request-permissions-at-runtime"
+                - title: "Add information about permissions to your extensions AMO page"
+                  id: "add-information-about-permissions-to-your-extensions-amo-page"
         - category: "User Experience"
           pages:
             - title: "Build a secure extension"
               url: "/documentation/develop/build-a-secure-extension/"
```
muffinresearch: This seems to be linking to accessibility documentation and the accessibility page is a 404 from the sidebar suggesting that the generation might not be quite right.

muffinresearch:
There's quite a few 404s from the sidebar nav, it would be a good idea to hookup some kind of link checker plugin to ensure we don't have broken links.

_includes/up-next.html
```diff
@@ -0,0 +1,216 @@
...
+{% assign max_steps = 3 %}
+
+<section class="module up-next">
+    <article class="module-content grid-x grid-padding-x">
+        <div class="cell small-12">
+            <h6>Up Next</h6>
+        </div>
+
+
+{% for item in site.data.pages %}
```
muffinresearch: This seems like quite a bit of duplication, I wonder if any of this could be factored out? I'm saying that without know what's possible via Jekyll templates though.

documentation/develop/build-a-secure-extension.md
```diff
@@ -0,0 +1,107 @@
...
+    "v=1&tid=" + GA_TRACKING_ID + "&cid= " + GA_CLIENT_ID + "&aip=1" +
+    "&ds=add-on&t=event&ec=AAA&ea=" + aType;
+
+  request.open("POST", "https://www.google-analytics.com/collect", true);
+  request.send(message);
+  {% endhighlight %}
+
+  You can find more information in the blog post [Using Google Analytics in Extensions](https://blog.mozilla.org/addons/2016/05/31/using-google-analytics-in-extensions/).
+
+- **Use the standard extension content security policy (CSP)**
```
muffinresearch: Could a better choice of markup be used here. The bold text preceding the rest of the copy doesn't appear to provide much meaning. Maybe something like a `dl` would be a better choice, albeit you'd probably be needing to express that in pure HTML.

Alternatively could this be headings and paras? It could be nice to be able to link to a specific point.

lancecummings:
> A gigantic 5k line patch is going to take some time to review. Please could you consider breaking up the changes further in future PRs?
> 
> As a yardstick a 500 line patch is around the max to aim for per branch, it's just a guide and sometimes that'll be exceeded, but more frequently posted smaller patches are a lot easier to review than one gigantic one that changes lots of different things.

Sorry about that @muffinresearch - I think you'll at least find that the vast majority of the changes don't need to be reviewed at this time, since the intent of this push is to get a starting point body of content into the Moz team hands to further review and adjust before publishing. Most of the images and content are just ported over from existing content hosted elsewhere.

lancecummings:
> Some of the images looks like illustrations so they might be better quality if svg was used rather than jpg:
> 
> * _assets/img/documentation/develop/EnterpriseExtensions_promo.jpg
> * _assets/img/documentation/develop/MultipleDevicesAndBrowsers_fullwidth.jpg
> * _assets/img/documentation/develop/PowerUpExtensions_fullwidth.jpg
> 
> Some of the images are rather large - please check and see if the optimization steps in the build process are going to be enough. There's a video background image that's 1920px wide, that probably doesn't need to be that big, and since the current image optimizations at build won't change the dimensions you'll need to check that the starting image is appropriately sized.

Oh yeah, I should have checked on how those ones were exported before adding them. I've done some further optimization on them now.

lancecummings:
> There's some 404s from the sidebar nav, it would be a good idea to hookup some kind of link checker plugin to ensure we don't have broken links.

Those are intentional at this point. The pages need to be built by the Moz team in the coming weeks.

community.md
lancecummings: Are you sure that a heading should be a direct child? My intent here was to have sections be flexible enough to contain an `article` and an `aside`, for instances when we have a related block of content in the right rail.

documentation/develop/build-a-secure-extension.md
lancecummings: This is content ported from this page: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Security_best_practices
It will be up to the content team to make content edits once they have access to these new pages. For the purposes of this pull request we should leave content revisions for later review.

community.md
muffinresearch: Certainly 2 outline checkers I used can't see a heading for the section element including: https://validator.w3.org/nu/

Showing_results_for_ew_html_-_Nu_Html_Checker

_includes/up-next.html
muffinresearch: How was this resolved?

_data/pages.yaml
muffinresearch: How was this resolved? Please add a comment so it's clear what action was taken.

muffinresearch:
> > There's some 404s from the sidebar nav, it would be a good idea to hookup some kind of link checker plugin to ensure we don't have broken links.
> 
> Those are intentional at this point. The pages need to be built by the Moz team in the coming weeks.

If you created placeholders and check the links this would ensure the navigation isn't broken like it was with the existing security and accessibility pages.

muffinresearch:
I think the sidebar navigation is going to need a re-think. It's not easy to understand the relationship between the headings in the way that you lose your place when you navigate to something lower down on the sidebar nav.

muffinresearch:
If there are no tags perhaps the tag ui shouldn't be shown (this is from /documentation/develop/)

Develop___Extension_Workshop

_includes/contents.html
```diff
@@ -0,0 +1,129 @@
+<div class="table-of-contents-inner">
+
+{% for item in site.data.pages %}
+
+    {% if item.categories %}   
+
+        {% if page.url == item.url %}
+            
+            {% if item.subpageitems %}
+                <h6>Contents</h6>
```
muffinresearch: There's still plenty of duplication here in this file.

_includes/up-next.html
```diff
@@ -0,0 +1,162 @@
...
+        {% for cat in item.categories %}
+
+
+                {% if cat.pages %}
+
+                        {% for pg in cat.pages %}
+
+                            {% if on_current_page == true %}
+
+                                {% assign next_page_ar = site.pages | where: 'permalink', pg.url %}
```
muffinresearch: There's still lots of duplicate logic throughout this file, can this be reduced further?

documentation/develop/build-an-accessible-extension.md
```diff
@@ -0,0 +1,237 @@
...
+topic: Develop
+tags: [Development, Extensions, UI, User Interface, UX, Web Extensions]
+contributors: [rebloor]
+last_updated_by: rebloor
+date: 2019-05-01 9:08:50
+---
+
+<!-- Page Hero Banner -->
+
+<section class="page-hero">
```
muffinresearch: Can all of these be includes too all with all the other "structural" markup used for the layout in each document. Since they're duplicated all over the place it would be a lot nicer to include them and have one instance to update in the event of future changes.

_data/pages.yaml
lancecummings: Ah, sorry! This page had the wrong front matter title and permalink, which I fixed.

_includes/up-next.html
lancecummings: I was able to use an include, so that the html only needed to be in one place.

community.md
lancecummings: I see. I guess we would ideally have a heading inside the section, in addition to a heading in the nested article. In our case, that would be redundant. I like how the section creates a container for the article/aside siblings, but would you suggest replacing these elements with divs?

_includes/contents.html
lancecummings: > If there are no tags perhaps the tag ui shouldn't be shown (this is from /documentation/develop/)
> 
> Develop___Extension_Workshop

If the tag isn't defined in the front matter, it won't show. I left an empty array in the front matter on those pages to highlight to the content team that they will need to fill in those tags (I believe that they are in the process of coming up with tag content for those pages)

documentation/develop/build-an-accessible-extension.md
lancecummings: Oh yeah, that's a very good idea. It would make the pages easier to read as well.

_includes/contents.html
muffinresearch: Ah ok, fair enough 👍

_includes/up-next.html
muffinresearch: I'll file a separate issue for further de-duplication.

PR #213, opened 2019-07-11
Sidebar scroll to active page

### LINKED ISSUE #209: Sidebar headings get hidden with a long side-bar upon navigation.
Having navigated to `/documentation/develop/` click on a page low down in the left-hand scroll-bar.

See the following: 

Build_a_secure_extension___Extension_Workshop

This example is not the worst case - it is possible for the side-bar high-lighted nav to be hidden completely.

There is also no bread-crumb navigation indicating where you are in the current hierarchy it just refers to `develop`, so the relationship between the current page and the other pages is not clear.

Build_a_secure_extension___Extension_Workshop

As a fix it would be good to consider:

* Limiting the side-bar to reflect the current place in the content hierarchy (e.g. having navigated to `/develop/build-a-secure-extension/` only the pages under the heading `User Experience` should be shown, ideally this would be accompanied with a breadcrumb update showing `Develop -> User Experience` to make it clear how to get back.
* The side-bar could be some kind of accordion that only shows the content under one sidebar heading at once but allowing opening the others. In the case above the content under `User Experience` would be visible but everywhere else would only show the headings. This still might have issues with enough content and a small enough viewport.

lancecummings:
Fixes #209

@muffinresearch - scrolls sidebar to active page on page load

### DIFF
--- a/_assets/js/main.js
+++ b/_assets/js/main.js
@@ -202,6 +202,13 @@ jQuery(document).ready(function($) {
     $('.up-next').hide(0);
   }
 
+  // 16. Scroll sidebar to active page position
+  // ------
+
+  if ($('.sidenav-detail .is-active').length) {
+    $('.sidenav').sidenavScrollToActive();
+  }
+
   // Init Breakpoint Listeners
   // ------------------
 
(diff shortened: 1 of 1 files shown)


muffinresearch:
This is excellent. Thanks for fixing this

Issue #248, opened 2019-07-26
Exact match of searched word placed lower than other results


mirunacurtean:
### Describe the problem and steps to reproduce it:

1. Open the Extension Workshop website: https://extensionworkshop-dev.allizom.org/. 
2. Click the “Search” button from the right side of the page.
3. Write “develop” in the search field and press Enter.
4. Observe the search results.

### What happened?
The exact match of the search is placed lower than other results.

### What did you expect to happen?
The exact match of the search should be placed higher than other results.

### Anything else we should know?
Reproduced with Fx68, Win 10x64 and Android 8.0.0. 

![exact search result placed lower](https://user-images.githubusercontent.com/53306802/61955016-d1ce8680-afc2-11e9-8898-729702510eb6.PNG)

caitmuenster:
@lancecummings, if there's enough time could you look into this?

lancecummings:
@caitmuenster this could be because currently we've got a boost level of 2 on pages that have a tag that matches a search term, and a boost level of 1 (lower boost) on page titles that match a search term. Would you like those boost settings adjusted?

caitmuenster:
@lancecummings yes, that would be great!

mirunacurtean:
Testing this fully is **Blocked** by issue https://github.com/mozilla/extension-workshop/issues/266. 
More precisely, for the "develop" and "publish" scenarios it could be verified that the exact match of the search word is displayed first, but it could not be confirmed for the other documentation title topics due to receiving 0 results when performing those searches. 
Also,on a separate note, when performing a search using the "extension" keyword, the first result was a link towards the "Enterprise" page instead of the "Extension Basics". Attached GIF for more details. 
![extension search results](https://user-images.githubusercontent.com/53306802/62116938-b6fe5980-b2c3-11e9-972d-b4b906344c60.gif)

PR #264, opened 2019-07-29
Boosting titles higher than tags

lancecummings:
@caitmuenster Boosting titles higher than tags and increasing boost of both tags and titles above body. Addresses issue #248


caitmuenster:
I'm still seeing titles coming up lower in search results: 

EW-search

lancecummings:
> I'm still seeing titles coming up lower in search results:
> 
> EW-search

Ok, I just boosted title to 25, and that seems to do the trick!

PR #265, opened 2019-07-29
escape key to close search in header

### LINKED ISSUE #251: Page header Search field is not dismissed after pressing ESC key

1. Open the Extension Workshop website: https://extensionworkshop-dev.allizom.org/.
2. Click the “Search” key from the right side of the page and the search bar will show. 
3. Press the ESC key.

### What happened?
The search field is not dismissed. 

The search field is not dismissed. 

Reproduced with Fx68, Win 10x64.
Only after clicking “x” the search field is dismissed. 
Reproduced also if clicking outside the search field.

lancecummings:
@caitmuenster this fixes #251

### DIFF
--- a/_assets/js/main.js
+++ b/_assets/js/main.js
@@ -1013,6 +1013,12 @@ jQuery(document).ready(function($) {
     $close.on('click', function() {
       $container.velocity('transition.slideRightOut', { duration: 450 });
     });
+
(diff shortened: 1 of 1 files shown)


caitmuenster:
This worked for me: 


ESC-confirm

PR #276, opened 2019-08-05
Added "Resources for Publishers" page & updated links

caitmuenster:
This PR adds a page for Resources for Publishers and updates links to the Documentation and Publish overviews.


caitmuenster:
To do - get rid of first commit.

Issue #249, opened 2019-07-26
Sub-string search not supported


vcarciu:
### Describe the problem and steps to reproduce it:

1.Open https://extensionworkshop-dev.allizom.org/
2.In searchbox , enter a substring like "communit"

Expected result:
We should display results for substrings

Actual results:
0 results , even if we have many results if we search for "community"

caitmuenster:
@lancecummings, if there's enough time could you look into this?

caitmuenster:
Actually -- @lancecummings, could you prioritize this?

caitmuenster:
Closed with #263.

vcarciu:
Substring search still not working as expected : 
![substring](https://user-images.githubusercontent.com/1583842/62146316-cdc1a200-b2fd-11e9-8eb3-56544f17ca46.png)

muffinresearch:
Ultimately Lunr has limited support for this. As was found with the patch in #263 it doesn't really work that well, it's really adding a wildcard to the entire query, this also turns off stemming.

As mentioned in https://github.com/mozilla/extension-workshop/pull/291#issue-308134374 if this is really a problem, then we might want to consider using a more fancy search engine or handing off to a 3rd party one, such as DDG because lunr definitely has some limitations and it's support for partial matches is quite limited and breaks quite easily.

caitmuenster:
After having some conversations, we're not going to make this issue a blocker. When we do the soft launch with the community, we're going to collect feedback on how search is performing and then we can reassess this.

PR #291, opened 2019-08-16
Revert partial search support

### LINKED ISSUE #266: Unable to find results using some of the documentation topic titles

1. Open the Extension Workshop website: https://extensionworkshop-dev.allizom.org/.
2. In either of the homepage search bars enter one of the following documentation topics titles : "Manage", "Enterprise", "Themes".  
3. View the search results. 

### What happened?
0 search results are found using the specified documentation topic titles. 
Also trying to perform a search for "community" will have the same result. 

Search results should be found for The documentations topics and also for other homepage titles like "Extensions basics" and "Community". 

Reproduced with Fx68, Win 10x64, Ubuntu 18.04.2 LTS 64-bit and Android 8.0.0. 
**Note: Results for the documentation topics titles can be found using a sub-string search, for example: "manag",  "enterpris" or "theme",  making this is to likely be a regression case from https://github.com/mozilla/extension-workshop/issues/249.** 

Search results are found only using substring

muffinresearch:
Fixes #266

Reverting this will cause partial matches to fail more, but even with this patch some partial strings don't work anyway see https://github.com/mozilla/extension-workshop/issues/249 which has been re-opened. 

Unfortunately Lunr has some limitations. If lunr.js is not good enough then maybe we should consider handing off to a proper search engine.


caitmuenster:
Apologies, @muffinresearch, it looks like there's a conflicting file. Would you be able to resolve and merge?

muffinresearch:
Landing https://github.com/mozilla/extension-workshop/pull/292 first includes this patch, so we can just close this out.

PR #302, opened 2019-08-20
Enable incremental builds for local dev

### LINKED ISSUE #300: Local builds are really slow
From @muffinresearch at https://github.com/mozilla/extension-workshop/pull/297#issuecomment-522676472: 

> The local builds are only slow when starting from scratch. After you've done it once thereafter a cache is used which speeds it up.

 > However, I suspect that some of the includes are being used heavily across the entire site and the build is probably executing that code over and over again. To fix that we should investigate profiling the build and see if anything sticks out as being particularly slow or accessed a lot of times.

muffinresearch:
Fixes #300 (for builds after the first one at least.)

See https://github.com/mozilla/extension-workshop/issues/300#issuecomment-523013477 for details

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "test": "npm run prettier-ci && npm run lint && npm run stylelint",
     "clean": "rimraf './.jekyll-cache/' && jekyll clean",
     "build": "JEKYLL_ENV=production SVGO_BIN='node_modules/svgo/bin/svgo' jekyll build",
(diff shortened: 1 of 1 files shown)


andrewtruongmoz:
Enabling incremental builds resulted in new content updates to the sidebar (pages.yaml) and "up next" at the bottom of the pages to not show on my local build. Removing --incremental from package.json solved the issue.

PR #335, opened 2019-08-30
Update the .gitignore file

### LINKED ISSUE #334: Update the .gitignore file
We should increase the list in .gitignore to ignore the files like `.DS_Store`. For example added in #304 by mistake.

championshuttler:
Fixes #334


.gitignore
```diff
@@ -6,3 +6,11 @@ node_modules/
 rvm-installer
 rvm-installer.asc
 *.log
+
+.DS_Store
+**/.DS_Store
+/.DS_Store
+__MACOSX/
+
+yarn.lock
```
muffinresearch: yarn.lock should not be ignored.

.gitignore
```diff
@@ -6,3 +6,11 @@ node_modules/
 rvm-installer
 rvm-installer.asc
 *.log
+
+.DS_Store
+**/.DS_Store
```
muffinresearch: Won't the previous line be enough?

.gitignore
```diff
@@ -6,3 +6,11 @@ node_modules/
...
+/.DS_Store
```
muffinresearch: I think this one is unnecessary too.

.gitignore
championshuttler: Oops my bad.

.gitignore
championshuttler: no actually in a PR i saw someone any how created the file in includes/ so thats why.

championshuttler:
> It would be great if you could also remove the`.DS_Store` files as part of this patch.

Already did ;)

.gitignore
muffinresearch: I add `.DS_Store` to master's `.gitignore` and then created a `.DS_Store` it's ignored. So I don't think you need the `**/.DS_Store` variant.

championshuttler: Thank you for investigation. Removed it :)

PR #429, opened 2019-09-23
Fix the resources link

### LINKED ISSUE #425: A resource linked in the Package your extension page is not available
1. Open the following section from the Package your extension page  - https://extensionworkshop-dev.allizom.org/documentation/publish/package-your-extension/#package-linux
2. Click on the following link from the information banner:

image

### What happened?
The linked page doesn't exist (a 404 is raised by the linked site)

The resource should be opened

@caitmuenster I'm not sure if we are using a wrong link or if the destination page is indeed not available anymore. In any case, this link should be updated.

championshuttler:
Fixes #425

### DIFF
--- a/documentation/publish/package-your-extension.md
+++ b/documentation/publish/package-your-extension.md
@@ -145,7 +145,7 @@ The most convenient way to package your extension is to use [web-ext build](/doc
 
 {% capture note %}
 
-See [http://infozip.sourceforge.net/Zip.html](http://infozip.sourceforge.net/Zip.html)
+See [https://linux.die.net/man/1/zip](https://linux.die.net/man/1/zip)
 
 {% endcapture %}
 {% include modules/note.html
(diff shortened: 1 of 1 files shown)


muffinresearch:
See https://github.com/mozilla/extension-workshop/issues/425#issuecomment-534042006 for a closer alternative to what this link previously linked to.

documentation/publish/package-your-extension.md
```diff
@@ -145,7 +145,7 @@ The most convenient way to package your extension is to use [web-ext build](/doc
 
 {% capture note %}
 
-See [http://www.info-zip.org/mans/zip.html](http://www.info-zip.org/mans/zip.html)
+See [https://linux.die.net/man/1/zip](https://linux.die.net/man/1/zip)
```
muffinresearch: Some copy for the link might make it a bit clearer what it's linking to e.g something like:

```markdown
See [the documentation for the `zip` command](https://linux.die.net/man/1/zip) for more information.
```

Renders as:

See [the documentation for the `zip` command](https://linux.die.net/man/1/zip) for more information.

PR #457, opened 2019-09-25
Fire media changes when matches are false as well as true

### LINKED ISSUE #433: [Android] Menu and dropdown become inactive when changing orientation
STR:
1. Load https://extensionworkshop-dev.allizom.org/ on a tablet in it's original orientation (landscape).
2. Go to a tab - extension Basics or any other
3. Change orientation into portrait
4. Tap the Menu button or the drop down bar (where the links from the pages are available)

Expected result:
The Menu is displayed. The drop down list is available. 

Actual result:
Both buttons are inactive.

Notes:
This issue is reproducible on tablets. I tested with a Samsung and a Huawei tablet both with Android 8.0. 
It is reproducible on all pages (including Homepage). 
I could not reproduce it with small devices - phones. 

Menu is inactive

As a workaround: if the page is refreshed the buttons will work again.

muffinresearch:
Fixes #433

The breakpoint events were only published when matches for a subscribed media query were true. This change makes it fire when matches is not undefined instead so it will fire for both true and false. 

The way the breakpoints code is written this means it will fire more often because it fires for every media query defined in breakpoints.js, however, the subscribed events seem to be no-op if the change doesn't need to be made, so this shouldn't cause a problem.

At some point in the breakpoints code could be refactored further.

### DIFF
--- a/_assets/js/breakpoints.js
+++ b/_assets/js/breakpoints.js
@@ -14,6 +14,7 @@
  * Changelog:
  * v 1.0 20.09.2013
  * v 2.0 28.02.2014 -- added configurable breakpoints
+ * v 2.1 25.09.2019 -- fixed breakpoints to publish if matches is defined.
  *
  * --------------------------------------------------------------------
  */
@@ -55,7 +56,7 @@
     // media query change
     // ------------------
     function breakpoint(mq) {
-      if (mq.matches) {
+      if (typeof mq.matches !== 'undefined') {
         var media = {};
         for (var i = 0; i < matches.length; i++) {
           media[matches[i].label] = matches[i].match.matches;


_assets/js/breakpoints.js
```diff
@@ -55,7 +56,7 @@
     // media query change
     // ------------------
     function breakpoint(mq) {
-      if (mq.matches) {
+      if (typeof mq.matches !== 'undefined') {
```
kumar303: This is the only place I see `breakpoint()` ever called:

https://github.com/mozilla/extension-workshop/blob/aa3b52edd9b8189972c4ff3202bd7fbcca1764d6/_assets/js/breakpoints.js#L51

Where does it get called with `{ matches: false }`?

muffinresearch: Those happen based on the event added here: https://github.com/mozilla/extension-workshop/blob/b33a10fe2d04de8fe49ab52b3a8541f07addc7d3/_assets/js/breakpoints.js#L47

`breakpoint` gets called with change events as a result of that.

PR #483, opened 2019-10-02
Remove scroll effects for touch devices

### LINKED ISSUE #413: There's a glitch on the screen when scrolling down towards the footer of the page
STR:
1.  Load https://extensionworkshop-dev.allizom.org/ on a mobile device and navigate toward the end of the page to `Connect with us` and `Footer`
2. Scroll down then up and repeat some scrolls up and down on those sections.
3. Observe the bottom of the screen 

Expected result:
No UX issues.

Actual result:
There's a white line (sometimes is more obvious) that appears on the bottom of the screen.

Notes:
This issue is reproducible with a Pixel and a Samsung phone and a Huawei Tablet all operating on Android 8.0.0.
The glitch is reproducible on all pages since the sections are available everywhere. 

Pixel with Android 8.0.0

Pixel

Huawei with Android 8.0.0

HUaweiMediaPad

### LINKED ISSUE #481: [Android] There is a glitch in the homepage footer that affects interaction with footer elements
1. Open Extension Workshop homepage on an Android device 
2. Go down to the footer section
3. Try to scroll up and down while in the footer area and observe its behavior 

### What happened?
The footer is jumpy, some elements might disappear from view.  You might also notice that you can't interact with elements (see the email field in the newsletter section)

The homepage footer behaves correctly 

- I did not reproduce the issue in other EW pages
- I've also checked with Chrome where the jumpiness is even more pronounced (see second screencast)
- reproduced with Fx68 and Chrome, Android 8.0 - all EW servers

[Firefox] - I've tried to capture the part where elements can't be interacted with as expected

[Chrome]

muffinresearch:
Fixes #481

This may also help #413 

## Update

This patch now only disables parallax effects for touch devices.

## Original description

This patch disabled scroll based events from working on touch devices. This means things like the menu are fixed to the top of the viewport irrespective of scrolling, trying this on a small screen seems to work ok, and it's not preventing too much of the content from being seen.

### Downsides

Landscape scrolling on a small screen isn't as good with the menus in the way.

Screen Shot 2019-10-02 at 13 49 06

Worst case with the browser chrome this looks like this:

IMG_6177

Which is really limited space to read.

### DIFF
--- a/_assets/js/main.js
+++ b/_assets/js/main.js
@@ -84,19 +84,21 @@ jQuery(document).ready(function($) {
   // **    the plugin code is found in parallax.js     **
   // *****                                         ******
 
-  if ($('.parallax').length) {
-    $('.parallax').parallax({ offsetIntertia: -0.15 });
+  if (typeof window !== 'undefined' && !('ontouchstart' in window)) {
+    console.debug('Parallax effects enabled');
+    if ($('.parallax').length) {
+      $('.parallax').parallax({ offsetIntertia: -0.15 });
+    }
+    if ($('.parallaxFG').length) {
+      $('.parallaxFG').parallaxFG({ offsetIntertia: 0.15 });
+    }
+    if ($('.parallaxFG-right').length) {
+      $('.parallaxFG-right').parallaxFG({ offsetIntertia: 0.075, axis: 'x' });
+    }
+    if ($('.parallaxFG-left').length) {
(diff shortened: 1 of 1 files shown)


muffinresearch:
Because of the space restrictions (see Downsides) I'm going to see if first just disabling the parallax effects when they're not needed is enough first, and further scroll effects can be disabled if needed.

PR #485, opened 2019-10-03
Fix trademark angle brackets

caitmuenster:
Markdown was swallowing angle brackets around "Add-on name"


caitmuenster:
cc @wagnerand

PR #486, opened 2019-10-03
Added missing "unexpected features" copy to No Surprises

caitmuenster:
We were missing a few bullet points from the No Surprises policy in the copyover from MDN.


caitmuenster:
cc @wagnerand

PR #489, opened 2019-10-04
Refactor the nav styles for touch-based devices

### LINKED ISSUE #481: [Android] There is a glitch in the homepage footer that affects interaction with footer elements

muffinresearch:
Fixes #481

This patch does a couple of things:

* Turns off the scroll based nav events for touch-enabled devices.
* Modifies the CSS so that position sticky is used for touch devices.

The results of this mean that the nav shows when scrolling up for non-touch devices as it did before. But for touch devices it uses sticky positioning to provide the most relevant nav at the top of the viewport during scroll. This is a compromise to avoid the issues as noted in "Downsides" here https://github.com/mozilla/extension-workshop/pull/483

### Notes for testing:

Either in a local branch or when this gets to -dev you can see the difference between touch and not-touch devices purely using the web-dev tools. Switch this icon and reload to see the differences.

Naturally this will also need to be tested on Android to ensure that the removal of scroll events helps the performance as expected. 

Community___Extension_Workshop

### DIFF
--- a/_assets/css/_sidebar.scss
+++ b/_assets/css/_sidebar.scss
@@ -6,10 +6,6 @@ Stylesheet: Side Bar Page Stylesheet
 
 // Template Body
 
-.site.sidebar {
-  background-color: $lighter-gray;
-}
-
 // Wrapper
 
 .site.sidebar .site-wrapper {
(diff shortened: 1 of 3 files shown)


_assets/css/_sidebar.scss
```diff
@@ -249,7 +268,7 @@ $sidebar-nav-width: rem-calc(240);
 // Side Nav Detail Loop
 
 .site.sidebar .sidenav-detail {
-  padding-bottom: rem-calc(16);
+  padding-bottom: rem-calc(64);
```
willdurand: I am not sure how this works but there are a few values that are repeated. We could maybe add variables to improve readibility.

muffinresearch: Yeah, vars sounds good, we'd need to know what they represent though.

_assets/css/_sidebar.scss
```diff
@@ -33,82 +29,106 @@ Stylesheet: Side Bar Page Stylesheet
...
-          display: none;
-        }
+    .is-active,
+    .has-active-children {
+      > a:hover {
+        text-decoration: none;
+      }
+    }
+
+    .primary > li:not(.is-active):not(.has-active-children) {
```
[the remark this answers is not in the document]
muffinresearch: _Note: a lot of this is just moved._ 

/me hides.

Issue #405, opened 2019-09-16
The popup-panels from Extension anatomy section are not dismissed when scrolling out of view


AlexandraMoga:
### Describe the problem and steps to reproduce it:

1. Open EW homepage - https://extensionworkshop-dev.allizom.org
2. Scroll down until you reach the Anatomy of an Extension section
3. Click on one of the tiles from the graphic 
4. After the popup-panel is opened, scroll down

### What happened?
The popup-panel is not dismissed

### What did you expect to happen?
The popup-panel should be dismissed

![2019-09-16_14-09-40](https://user-images.githubusercontent.com/31961530/64956906-5be4ee00-d894-11e9-9186-ef90fd7e5339.gif)

### Notes:
Reproduced with Fx69, Win10x64

muffinresearch:
It should also be noted the in-view code is disabled for touch devices in #504 to fix scroll perf/jank on Android. I'm beginning to think the right fix would be to make the overlays stay relative to the content that spawns them rather than try and close them based on scrolling of any kind. Tying into scrolling in any way seems to have historically causes a lot of undesirable issues on android.

PR #492, opened 2019-10-05
fix: [Issue #405] Add scroll event to close popup panels

ajaguasin:
Hi, this is my first time contributing.

I wasn't sure where to add my username and the date for each file I changed.

Please let me know, and i'll adjust accordingly,

Thanks!

### DIFF
--- a/_assets/js/main.js
+++ b/_assets/js/main.js
@@ -875,6 +875,10 @@ jQuery(document).ready(function($) {
             if (e.key == 'Escape' && $panel.is(':visible')) {
               closePanel($link, $panel);
             }
+          })
+          .on('scroll', function() {
+            closePanel($link, $panel);
+            $(document).off('scroll');
           });
       }
     }


_assets/js/main.js
```diff
@@ -875,6 +875,10 @@ jQuery(document).ready(function($) {
             if (e.key == 'Escape' && $panel.is(':visible')) {
               closePanel($link, $panel);
             }
+          })
+          .on('scroll', function() {
```
muffinresearch: This event should be namespaced - see above for examples. It should also not operate if there's no panel open.

_assets/js/main.js
```diff
@@ -875,6 +875,10 @@ jQuery(document).ready(function($) {
...
+            closePanel($link, $panel);
+            $(document).off('scroll');
```
muffinresearch: This should probably be called in `closePanel` instead (don't forget to use the namespace).

muffinresearch:
On reflection, I'm wondering if scroll is the right way to do this, really this should fire when the digram is no longer in-view. Since the diagram's animations are tied to being in view, that might provide a better hook rather than simply listening to scroll. The problem with scroll is that as soon as you touch the mouse the overlay will be closed, that may create an accessibility issue.

_assets/js/main.js
muffinresearch: See my latest comment, I think scroll events might not be the best hook for this.

ajaguasin:
@muffinresearch Thanks for the feedback. I'll investigate it more and update the PR

muffinresearch:
See also https://github.com/mozilla/extension-workshop/issues/405#issuecomment-540973143

muffinresearch:
@ajaguasin hi, since this hasn't seen updates for a while I'm going to close this, please feel free to re-open it when it's ready for another look.

PR #508, opened 2019-10-13
fix:503 Landscape with browser chrome makes heading cramped on the ho…

### LINKED ISSUE #503: Landscape with browser chrome makes heading cramped on the homepage
Using an Nexus 7 (Android) running FF Android, when the browser chrome is open it looks like this:

Screen Shot 2019-10-10 at 16 43 12

We should add more space before the heading to allow for this.

mukul-kmr-jha:
Fixes #503

@muffinresearch  Please take a look , and let me know if anything needs to be changed.
Thanks!

### DIFF
--- a/_assets/css/_fullwidth.scss
+++ b/_assets/css/_fullwidth.scss
@@ -88,6 +88,16 @@ Stylesheet: Full Width Page Stylesheet
     height: 86%;
 
     .cell {
+      margin-top:2.33rem;
+      font-size:0.85rem;
+      @include breakpoint(medium ) {
+        margin-top:6.5rem;
+        font-size:1rem;
+      }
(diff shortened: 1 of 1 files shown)


muffinresearch:
Doesn't appear to work from trying this out:

Extension_Workshop___Firefox_Extension_Workshop__Make_the_web_do_what_you_want__If_you_can_dream_it__our_open_souce_API_and_your_HTML__CSS__and_JavaScript_skills_can_make_it_happen_

Modifying `.cell` might be too generic, _hint: maybe look changing the `height` on `.banner.intro .banner-content`_

muffinresearch:
Thanks for the patch thus far, however I'm going to close this in favour of #510 and some modifications which was a bit closer to a fix.

PR #514, opened 2019-10-25
Only show tags/contributors when we have some

willdurand:
Fixes https://github.com/mozilla/extension-workshop/issues/515

---

Before:

Screen Shot 2019-10-25 at 12 42 29

After:

Screen Shot 2019-10-25 at 12 42 17


muffinresearch:
Thanks for the fix @willdurand

PR #520, opened 2019-10-31
fix: update to latest Firefox Browser branding

coogie:


muffinresearch:
@caitmuenster thinking this might need a review from the branding team - that said though if the workmark fonts are right maybe we can simply land this without changing anything else?

caitmuenster:
Thanks for the patch, @coogie! 😍 It's really nice to have help updating branding. :) If we're going to update the favicon, we should probably update the logo that's used on the header menu bar: 

Screen Shot 2019-11-01 at 12 27 22 PM

I'm also not sure if we need official brand sign-off to land this. @aaronrbenson, what are your thoughts?

muffinresearch:
> Thanks for the patch, @coogie! 😍 It's really nice to have help updating branding. :) If we're going to update the favicon, we should probably update the logo that's used on the header menu bar:
> 

This is just caching, unfortunately sometimes jekyll or jekyll-assets fails to see the changed files. 

Extension_Workshop___Firefox_Extension_Workshop__Make_the_web_do_what_you_want__If_you_can_dream_it__our_open_souce_API_and_your_HTML__CSS__and_JavaScript_skills_can_make_it_happen_

caitmuenster:
*shakes fist at caching* Thanks @muffinresearch! :)

muffinresearch:
Patch looks good to me, though we could shave a few bytes running the images though optimizations. I can do that as a follow-up, so if it works to land it to give @aaronrbenson a closer look that's fine by me.

coogie:
Thanks, @caitmuenster and @muffinresearch, glad I can contribute - I owe most of my career to Moz and the people (such as yourselves) behind it, so it's the least I can do 💖

WRT image optimisation, I gave them a cursory run through https://squoosh.app/, but there might be more that bytes that can be shorn off via alternative methods; happy to update accordingly.

aaronrbenson:
Looks great to me!

PR #525, opened 2019-11-06
Update to note deprecation of sideloading

jvillalobos:
Ref: https://blog.mozilla.org/addons/2019/10/31/firefox-to-discontinue-sideloaded-extensions/

This change would work best in tandem with the sideloading and registry install articles having a deprecation notice up. I also don't know what docs we have on Enterprise installation.


caitmuenster:
Thanks @jorgev. We have two pages for Enterprise that might be impacted: 
- https://extensionworkshop.com/documentation/enterprise/enterprise-distribution/
- https://extensionworkshop.com/documentation/enterprise/enterprise-policies-that-impact-extensions/

PR #526, opened 2019-11-07
Sideloading deprecation

caitmuenster:
Adds deprecation notice to `documentation/enterprise/enterprise-distribution/#installation-using-windows-registry` and `/documentation/publish/distribute-sideloading/#standard-extension-folders`. 

@jvillalobos, could you review this?


jvillalobos:
Suggested an edit on Slack, otherwise lgtm

PR #570, opened 2020-02-28
Fixed a minor spelling

grlwholifts:
- Changed spelling from 'remembear' to 'remember'


caitmuenster:
Thanks! :) In this case, the misspelling is ok because "RememBear" is the name of the extension. 😉

PR #649, opened 2020-06-14
docs: add --args and fix headings

lfilho:
`--args` was added a few months back but not yet reflected on the workshop site.

This would have saved me a couple of hours of digging around the web :-)

Also fixed the headings of two other subsections insider `run` to be at the right level.

### DIFF
--- a/documentation/develop/web-ext-command-reference.md
+++ b/documentation/develop/web-ext-command-reference.md
@@ -7,6 +7,7 @@ topic: Develop
 tags: [commands, Options, Reference, Tools, web-ext, WebExtensions]
 contributors:
   [
+    lfilho
     rebloor,
     mdnwebdocs-bot,
     kumar303,
@@ -23,8 +24,8 @@ contributors:
     sharang,
     chrisdavidmills,
   ]
-last_updated_by: rebloor
-date: 2019-04-16 10:25:47
+last_updated_by: lfilho
+date: 2020-06-14 00:11:47
 ---
 
 <!-- Page Hero Banner -->
@@ -286,11 +287,18 @@ If no target is specified, the extension will run in `firefox-desktop`.
 
 Environment variable: `$WEB_EXT_TARGET`
 
-### \--chromium-binary
+#### \--args, --arg
+Additional CLI options passed to the Browser binary. Example:
+
+{% highlight javascript %}
+--arg "'--search=mozilla'" --arg "'--new-tab=https://duckduckgo.com'"
+{% endhighlight %}
+
+#### \--chromium-binary
 Path or alias to a Chromium executable such as google-chrome, google-chrome.exe or opera.exe etc.
 If not specified, the default Google Chrome will be used.
 
-### \--chromium-profile
+#### \--chromium-profile
 Path to a custom Chromium profile.
 
 #### \--start-url
@@ -560,3 +568,4 @@ WEB_EXT_VERBOSE=true
 {%- include up-next.html -%}
 
 <!-- END: Up Next -->
+
--- a/documentation/develop/web-ext-command-reference.md
+++ b/documentation/develop/web-ext-command-reference.md
@@ -7,7 +7,7 @@ topic: Develop
 tags: [commands, Options, Reference, Tools, web-ext, WebExtensions]
 contributors:
   [
-    lfilho
+    lfilho,
     rebloor,
     mdnwebdocs-bot,
     kumar303,
--- a/documentation/develop/web-ext-command-reference.md
+++ b/documentation/develop/web-ext-command-reference.md
@@ -291,7 +291,7 @@ Environment variable: `$WEB_EXT_TARGET`
 Additional CLI options passed to the Browser binary. Example:
 
 {% highlight javascript %}
---arg "'--search=mozilla'" --arg "'--new-tab=https://duckduckgo.com'"
+--arg="--search=mozilla" --arg="--new-tab=https://duckduckgo.com"
 {% endhighlight %}
 
 #### \--chromium-binary


caitmuenster:
Thanks for the patch, @lfilho! 

@rpl and @Rob--W, could you review?

lfilho:
Oh, thanks for catching my typo :)

documentation/develop/web-ext-command-reference.md
```diff
@@ -560,3 +568,4 @@ WEB_EXT_VERBOSE=true
 {%- include up-next.html -%}
 
 <!-- END: Up Next -->
+
```
rpl, line 571: Nit, the newline added to the end of the file is likely not needed, but not a big deal.

documentation/develop/web-ext-command-reference.md
```diff
@@ -286,11 +287,18 @@ If no target is specified, the extension will run in `firefox-desktop`.
...
+{% highlight javascript %}
+--arg "'--search=mozilla'" --arg "'--new-tab=https://duckduckgo.com'"
+{% endhighlight %}
+
+#### \--chromium-binary
 Path or alias to a Chromium executable such as google-chrome, google-chrome.exe or opera.exe etc.
 If not specified, the default Google Chrome will be used.
 
-### \--chromium-profile
+#### \--chromium-profile
```
rpl, line 301: Thanks for fixing the heading level on those two entries!

documentation/develop/web-ext-command-reference.md
```diff
@@ -286,11 +287,18 @@ If no target is specified, the extension will run in `firefox-desktop`.
 
 Environment variable: `$WEB_EXT_TARGET`
 
-### \--chromium-binary
+#### \--args, --arg
+Additional CLI options passed to the Browser binary. Example:
+
+{% highlight javascript %}
+--arg "'--search=mozilla'" --arg "'--new-tab=https://duckduckgo.com'"
```
rpl, line 294: @lfilho  I think that the double-quoting (using both `"` and `'`) here on the --arg values is currently needed to workaround this yargs related web-ext issue mozilla/web-ext#1846, is that right?

That should be fixed once we release web-ext 4.3.0 (fixed by updating yargs see mozilla/web-ext#1902)

I think that is reasonable that the example here does work in the version of web-ext currently released (4.2.0) as it is, but maybe we can remove the double quotes once web-ext 4.3.0 has been released.

Rob--W: Double-quoting shouldn't be needed in either case. The quotes only matter for the shell.

@rpl Even with the yargs bug, it can be worked around by `--arg=\ --search=mozilla`, for example.

I would go with `--arg=--search=mozilla` instead of `--arg --search=mozilla`, for clarity.

caitmuenster: Thanks @Rob--W and @rpl! @lfilho, want to update the patch with Rob's last comment? Then we'll approve + merge. :)

documentation/develop/web-ext-command-reference.md
lfilho: Sorry, auto-added by my editor. Maybe a case for [`.editorconfig`](https://editorconfig.org/)? It has a config for that.

documentation/develop/web-ext-command-reference.md
lfilho: Sorry, @Rob--W, your suggestion doesn't work. It seems to think `--search` as an arg to web-ext itself instead of passing it to the browser:

```
> web-ext run --args=--search=mozilla --browser-console --url ./__tests__/shared/test_page.html

web-ext run

Run the extension

Options:
  --version                        Show version number                 [boolean]
[...]
[...]

Unknown argument: search
```

Rob--W: It's `--args=\ --search=mozilla` with `=`. But I only showed it to show that quotes aren't needed.

The second issue, `--args=--search=mozilla` doesn't work on the current version (4.2.0), but will work with the upcoming release, `web-ext` 4.3.0. We intend to release this version this week.

rpl: Personally I think that the `--args=\ --search=mozilla` workaround that Rob did suggest risks to be even more confusing that the double-quotes.

@lfilho @Rob--W I briefly discussed about this PR with @caitmuenster today and agreed to proceed with the following steps:

- wait for the web-ext v4.3.0 release to get on npm (I just tagged the new v.4.3.0 release and the travis job should release it on npm as soon as all travis tasks did pass)
- then let @lfilho to try and confirm that mozilla/web-ext#1846 is fixed in web-ext v4.3.0 as expected (and that an example using `--args="--search=mozilla"` would work as expected)
- then update this PR to not using the double-quotes (nor the `--args=\ --search=mozilla` workaround that Rob did suggest, personally I think that it risks to be even more confusing that the double-quotes) and merge it on master

@lfilho @Rob--W @caitmuenster how that sounds?

lfilho: Sounds good to me. I'll do it over the weekend.

caitmuenster: Thanks so much @lfilho! FYI I will be out on PTO from June 25 - July 6. If your patch doesn't get reviewed and merged by someone else during that time I will do it when I get back. :)

lfilho: The below did work:

- ./node_modules/web-ext/bin/web-ext run --verbose --args="--search=mozilla"
- ./node_modules/web-ext/bin/web-ext run --verbose --args=--search=mozilla
- ./node_modules/web-ext/bin/web-ext run --verbose --args=--search="mozilla firefox" 
- ./node_modules/web-ext/bin/web-ext run --verbose --args="--search=mozilla firefox" (this is weird, should it?)

The below did NOT work:
- ./node_modules/web-ext/bin/web-ext run --verbose --args "--search=mozilla firefox" (based on your comments above this is expected, but most of the other args in this documentation use this format, so it's weird that for `args` we **must** use `=` version.

lfilho: Just published a new commit as agreed :)

PR #655, opened 2020-07-08
Fix link to eval doc

rugk:
It seems the heading has changed is now even more anti-eval. 😜


muffinresearch:
Thanks this was landed in https://github.com/mozilla/extension-workshop/commit/6595dca1874eeaab670e1ae31463352c45f9e4ea

Issue #584, opened 2020-03-09
Shorten all other page titles to less than 70 characters 


caitmuenster:
Note: this issue is probably not appropriate for contributors at this time. 

Shortening all page titles to less than 70 characters will help with SEO. This is related to #582. 

Before this issue can be fixed, we probably need to supply the 70 character or less titles for all pages. 

I think this can be fixed by updating the 'title' line in each .md file: 

![Screen Shot 2020-03-09 at 10 23 10 AM](https://user-images.githubusercontent.com/23342476/76240172-04df2680-61f0-11ea-8c2b-a7c39a9030b6.png)

@muffinresearch, is that right?

muffinresearch:
This should have a test that looks at the yaml and validates it when this is fixed.

muffinresearch:
The test added in #680 covers this from the perspective of the titles coming from frontmatter. I'm not sure we need to go much further than this.

https://github.com/mozilla/extension-workshop/blob/445670542d9e2da99f46f7aeb765a247322539f6/tests/frontmatter.test.js#L8-L14

PR #680, opened 2020-07-24
Add some basic initial tests

muffinresearch:
Relates to #584

This adds jest and some very simple tests looking at the frontmatter.


package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "dompurify": "2.0.12",
     "foundation-sites": "6.6.3",
+    "glob": "^7.1.6",
```
willdurand, line 18: Should we get rid of the carets in the added deps?

tests/frontmatter.test.js
```diff
@@ -0,0 +1,28 @@
...
+    let fm = matter.read(mdFile).data;
+
+    // Ignore the homepage for title and description as it's special.
+    if (!mdFile.match(/index\.md/)) {
+      it(`${mdFile} should have description under 70 chars`, () => {
+        expect(fm.title).toBeDefined();
+        expect(fm.title.length).toBeLessThanOrEqual(70);
+      });
+
+      // Skipping to start with because we don't have this yet.
```
willdurand, line 15: Then maybe remove it?

tests/frontmatter.test.js
```diff
@@ -0,0 +1,28 @@
+const glob = require('glob');
+const matter = require('gray-matter');
+
+describe('frontmatter tests', () => {
+  for (const mdFile of glob.sync('src/content/**/*.md')) {
```
willdurand, line 5: Did you try to use jest `.each` instead?

muffinresearch: I did look at it, but it's not a case of re-using the same data each time, the expected vs input is different. I think overall this approach is more direct and easier to understand.

tests/frontmatter.test.js
muffinresearch: Yep, fair enough - I'll add this example to the issue for adding descriptions as a guide.

tests/frontmatter.test.js
willdurand: ok yep, works for me in any case 👍

muffinresearch:
@willdurand Thanks for the review!

PR #722, opened 2020-08-07
Add tests for asset pipeline

muffinresearch:
This patch adds some tests for the asset-pipeline script which will help a lot for adding new features and any additional refactoring.


muffinresearch:
Discussed this with @eviljeff and he's prioritizing other reviews currently. As this is the static build script there's not a huge amount that can go wrong. So I'm going to land it and check that it's working on -dev before promoting it to stage.

PR #766, opened 2020-09-16
Update about DOMPurify 

caitmuenster:
* Adds note and warning about recommended DOMPurify versions for "Build a Secure Extension" tips.


caitmuenster:
@wagnerand, r?

wagnerand (approved):
Looks good, thank you. Do we also want to add a note that only the latest versions will be accepted by AMO? (ref https://github.com/mozilla/addons-linter/blob/master/docs/third-party-libraries.md)

caitmuenster:
@wagnerand Added! Thanks!

PR #786, opened 2020-09-30
Update Firefox for Android article notes

caitmuenster:
Add a link to the Add-ons Blog for how to persistently install extensions on Firefox for Android Nightly. 

@muffinresearch if you could give this a quick once-over that would be great. :)


muffinresearch (approved):
LGTM r+

Issue #673, opened 2020-07-21
Use markdown-it plugin for toc


muffinresearch:
Instead of using page data to build the toc it would be better to fix https://github.com/mozilla/extension-workshop/issues/659 and then use a toc plugin.

The same slugify function should be used everywhere for ids as that will make everything consistent:

Here's a sample of code that does the toc and heading id generation:

```js 
    [require('markdown-it-anchor'), {
      permalink: false,
      slugify: slugify
    }],
    [require('markdown-it-toc-done-right'), {
      listType: 'ol',
      level: 2,
      slugify: slugify,
      containerClass: 'table-of-contents-inner',
      callback: (html, ast) => {
        html = `<aside class="module-aside table-of-contents">${html}</aside>`;
      },
    }],
```

This should also then negate the need for doc authors working around issues with this by adding `<section id="some-id-just-for-linkage">`.

Issue #659, opened 2020-07-10
Refactor nav to use 11ty navigation


muffinresearch:
Ideally to add a new page to the docs site you'd just have to to add a single page in markdown, not add data to a JSON file.

It would be good to take a look at https://www.11ty.dev/docs/plugins/navigation/ and see if that could be used to generate the necessary data for building the nav from frontmatter alone.

PR #800, opened 2020-10-11
docs: fix titles

noraj:
unwanted HTML section tags were breaking the markdown parsing

image

### DIFF
--- a/src/content/documentation/develop/web-ext-command-reference.md
+++ b/src/content/documentation/develop/web-ext-command-reference.md
@@ -22,9 +22,10 @@ contributors:
     tofumatt,
     sharang,
     chrisdavidmills,
+    noraj,
   ]
-last_updated_by: lfilho
-date: 2020-06-14 00:11:47
+last_updated_by: noraj
+date: 2020-10-11 18:11:47
 ---
 
 <!-- Page Hero Banner -->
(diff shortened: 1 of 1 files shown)


muffinresearch:
As horrible as these sections are, I think the ids are needed (unless it's clear they're not linked to) until this can be properly refactored so that ids are correctly auto-generated and hacks like this are no longer required. As a quick fix, adding a blank line of space under the HTML should fix the rendering. 

A div would almost certainly be a better choice for this rather than having meaningless empty sections, but at some point this whole thing needs replacing so that this kind of thing is no longer needed. See also: #659 #673

noraj:
@muffinresearch Yeah using markdown-it would be what I recommend too (#673). Yeah I could update this PR to add a blank line but is it even wort it? Maybe I'll close this PR in favor for #673 .

muffinresearch:
Adding newlines would at least un-break this page until #673 can be fixed, which is going to be a way off since it's not just as simple as adding the plugin (mainly because the toc is used on the sidebar, so that will make it more complicated)

muffinresearch:
@noraj thanks!

caitmuenster:
Thanks so much, @noraj! 🙌 We've added your contribution to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition).

PR #881, opened 2021-01-06
Update to contact address

mafeld:


wagnerand (approved):
This looks good to me, thanks! @mafeld is there a timeline, when this should go out? Any date before our after it should or should not be live?

mafeld:
Sometime next week would be good.

PR #894, opened 2021-01-25
Added newlines between headings and their anchors

Hwiet:
There are three headings not rendering correctly in the [published article](https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/) (rendering as paragraph text), I think this change should probably fix it.


muffinresearch:
Thanks @Hwiet for the patch. I'm going to land this along with a test in #896

muffinresearch:
Merged here: https://github.com/mozilla/extension-workshop/commit/b3345b7f6dc2a2f6b36144392589cc9951e7f05d

PR #895, opened 2021-01-25
Removes page for Promoted Add-ons Pilot Program.

caitmuenster:
Since we have wrapped up the [Promoted Add-ons Pilot](https://blog.mozilla.org/addons/2021/01/21/promoted-add-ons-pilot-wrap-up/), this PR removes the page about the program.


muffinresearch (changes requested):
Looks like this is referenced in `src/data/pages.json`. So that entry would need to also be removed.

muffinresearch (approved):
r+

PR #899, opened 2021-01-28
Replace Terser.parse with acorn

### LINKED ISSUE #704: Terser.parse is not a function
Terser 5 removed the access to the parse function. This now breaks the approach taken to use those function to rewrite the code. To fix this, I'm going downgrade to get it going then the `asset-pipeline` script really needs to be updated to use an alternative approach.

muffinresearch:
Fixes #704

Fixing this will allow Terser updates to resume.


willdurand (approved):
So if I understand correctly, we use `acorn` for parsing the JS and rewrite a few things, and then `terser` for minification only, right?

r+wc

muffinresearch:
> So if I understand correctly, we use `acorn` for parsing the JS and rewrite a few things, and then `terser` for minification only, right?
> 

Correct, this was only necessary because Terser removed access to it's internal AST parsing so they can make more widespread changes.

PR #910, opened 2021-02-09
Add documentation for `--watch-file`

### LINKED ISSUE #872: Missing Documentation for --watch-file command line option in extension workshop
Documentation for command line option `--watch-file` only exists when you run `web-ext run --help`

Dexterp37:
Fixes #872

This PR adds documentation for the `--watch-file` option (that was missing).

The behaviour for this option changed in mozilla/web-ext#2125.


ankushduacodes:
Heads up for the linked issue #872.

Just wanted to make a note of it so I can close it when this PR lands :)

ankushduacodes:
#### Question:
Shouldn't the `--watch-file` option now be called `--watch-files` instead to be more consistent with options like `--ignore-files`?

#### Suggestion:
`--watch-file` docs on `web-ext run --help` may also need some wording change to reflect the changes better.
https://github.com/mozilla/web-ext/blob/be8b8d3b8b967e88035826deb8141edd41c41b27/src/program.js#L619-L625


I apologise for not catching these while [#2125](https://github.com/mozilla/web-ext/pull/2125) was still under development. Just got around to it :-)

rpl:
> #### Question:
> 
> Shouldn't the `--watch-file` option now be called `--watch-files` instead to be more consistent with options like `--ignore-files`?

well, to be fair there are a number of other `type: 'array'` cli options that are also not plural (e.g. `--target`, `--watch-ignored`, `--pref`, `--start-url`).
In the end an option of type array doesn't have to be used for more than one value or more than one time, it is mainly "allowed to be used more than once" (also for yargs `--watch-file file1 file2` and `--watch-file file1 --watch-file file2` are equivalent and both valid), and so my personal opinion is that renaming the option isn't mandatory (especially given that the option did already exist before and the last change does just allow it to be used for more then just one watch file)

Alternatively we may use the same kind of approach we use for the `--arg / --args` option, and allow the users to use `--watch-files` as an alias:
https://github.com/mozilla/web-ext/blob/be8b8d3b8b967e88035826deb8141edd41c41b27/src/program.js#L664-L669

Still not strictly mandatory, but I would definitely be ok to make that change before release the new web-ext version. 

@ankushduacodes This doesn't mean that your point wasn't a good one to raise, I actually appreciate you chiming in and point that out!

> #### Suggestion:
> 
> `--watch-file` docs on `web-ext run --help` may also need some wording change to reflect the changes better.
> https://github.com/mozilla/web-ext/blob/be8b8d3b8b967e88035826deb8141edd41c41b27/src/program.js#L619-L625

This sounds reasonable, we can still tweak the description in a separate PR (and maybe also add the `watch-files` alias as well) before we are going to release web-ext new version.

> I apologise for not catching these while [#2125](https://github.com/mozilla/web-ext/pull/2125) was still under development. Just got around to it :-)

No need to apologies, and your timing is actually still pretty great, we do have time to follow up before this change is being released to the users.

src/content/documentation/develop/web-ext-command-reference.md
```diff
@@ -340,6 +340,14 @@ web-ext run --start-url www.mozilla.com --start-url developer.mozilla.org
 
 Environment variable: `$WEB_EXT_START_URL`
 
+#### `--watch-file`
```
rpl, line 343: Nit, `--watch-file`, `--watch-files` (because we added the alias in web-ext v.6.0.0, mozilla/web-ext#2182)

rpl (approved):
Just noticed that I didn't added the nit comment here after we merged that change on the web-ext side.

It is also totally fine to merge this as is and then add the alias in a separate PR, and so I'm also marking this version as approved.

PR #959, opened 2021-04-01
Update themes pages

### LINKED ISSUE #956: Move Theme Concepts page from MDN to EW
The [Theme Concepts](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/Themes/Theme_concepts) page on MDN is mostly Firefox-specific and should be moved into EW.

caitmuenster:
Fixes #956. 

This PR: 
* Creates new pages for static and dynamic theme content
* Updates content on all theme pages


src/data/pages.json
```diff
@@ -1328,23 +1328,75 @@
...
+                    "title": "Updating static themes",
+                    "url": "updating-static-themes"
+                  },
+                   {
+                     "title": "Single image themes",
+                     "url": "single-image-themes"
+                   },
+                   {
+                     "title": "Multiple image themes",
+                     "url": "multiple-image-themes"
```
muffinresearch, line 1371: I think all of these `"url"` keys should be `"id"` instead.

muffinresearch (changes requested):
Needs the json updating because otherwise the nav doesn't work correctly.

caitmuenster:
Thanks @muffinresearch! Most recent commit should fix the json.

muffinresearch (approved):
r+

PR #915, opened 2021-02-17
Duplicated links converted into references in markdown

Patil2099:
Fixes #241

PR #972, opened 2021-04-19
Duplicated links converted into references in markdown

### LINKED ISSUE #241: Links should be references in markdown
Rather than duplicate the same link many times in markdown, it would be much better to use a reference.

Here's an example:

Instead of 

## Original markdown

```markdown
* Something something [a link](https://mozilla.org/)
* Something else [a link](https://mozilla.org/)
```

We should use:

## Reference markdown

```markdown
* Something something [a link][duplicate link]
* Something else [a link][duplicate link]

[duplicate link]: https://mozilla.org/
```

Both render the same thing, but the reference solution makes updating that link much easier since there's only one!

## Original
* Something something [a link](https://mozilla.org/)
* Something else [a link](https://mozilla.org/)

## Reference version

* Something something [a link][duplicate link]
* Something else [a link][duplicate link]

[duplicate link]: https://mozilla.org/

patil2097:
Fixes #241

Sorry for the delay. Can you please review this @muffinresearch  and @caitmuenster ?


muffinresearch:
@patil2097 could you add changes to https://github.com/mozilla/extension-workshop/pull/915 instead of this new PR? It's helpful if we can keep the updates together in one place as that will save on cross-referencing.

PR #1011, opened 2021-05-13
Fix wrong utm_campaign parameter

rugk:
It’s likely a typo that you missed the ` utm_`  prefix there, which makes the mentioned example unusable. :slightly_smiling_face:


muffinresearch:
@rugk thanks for the patch.

PR #1051, opened 2021-07-05
Add viewing_log_output section at debugging

### LINKED ISSUE #1050: Document browser console at debugging (viewing_log_output)
The debugging section at https://extensionworkshop.com/documentation/develop/debugging/ does not mention the Browser Console anywhere. It should. The browser console sometimes contains more detailed error messages. In particular it also contains the output of stderr from native messaging (see https://bugzilla.mozilla.org/show_bug.cgi?id=1718691 for an example of a developer getting confused).

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging#troubleshooting currently links to https://extensionworkshop.com/documentation/develop/debugging/#viewing_log_output . We could add a new section to the extension workshop, which explains what I put here in more words and also link to https://developer.mozilla.org/en-US/docs/Tools/Browser_Console for even more details. In the past there was also a "viewing_log_output" anchor at the extension debugging page on MDN, at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Debugging_(before_Firefox_50)#viewing_log_output

Rob--W:
Fixes #1050.

### DIFF
--- a/src/content/documentation/develop/debugging.md
+++ b/src/content/documentation/develop/debugging.md
@@ -70,6 +70,9 @@ To debug your extension, you'll use:
 - [**Debugger**](https://developer.mozilla.org/docs/Tools/Debugger) to set breakpoints and watchpoints in your extension's JavaScript, and examine and modify the extension's state.
 - [**Storage**](https://developer.mozilla.org/en-US/docs/Tools/Storage_Inspector) to view details of data stored by the extension.
 
+This console shows messages from your background script and extension pages; other logs may appear elsewhere.
+There is an overview of all relevant places to look for messages at [Viewing log output](#viewing_log_output).
+
 For much of the debugging work, it's useful to be able to view **Console** with **Inspector** or **Debugger**. You do this using the [split console](https://developer.mozilla.org/docs/Tools/Web_Console/Split_console), press `esc` to activate this mode.
 
 ![developers tool split screenshot](/assets/img/documentation/develop/developing_tools_inspector_split.png)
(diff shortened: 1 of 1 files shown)


src/content/documentation/develop/debugging.md
```diff
@@ -70,6 +70,8 @@ To debug your extension, you'll use:
 - [**Debugger**](https://developer.mozilla.org/docs/Tools/Debugger) to set breakpoints and watchpoints in your extension's JavaScript, and examine and modify the extension's state.
 - [**Storage**](https://developer.mozilla.org/en-US/docs/Tools/Storage_Inspector) to view details of data stored by the extension.
 
+Some logs do not appear in the console of this Toolbox. There is an overview of all relevant places to look for messages at [Viewing log output](#viewing_log_output).
```
rpl, line 73: Nit, phrased as "Some logs do not" does sound a bit "mysterious" :-), maybe we may phrase it differently, something like `Be aware that not all logs appears in the console panel of this Toolbox, look to the section "Viewing log output" for an overview of all the relevant places to look for logs.`.

src/content/documentation/develop/debugging.md
```diff
@@ -90,6 +92,41 @@ You can now drag the toolbox tab to a separate window, so you can place it along
...
+  - [Debugging background scripts](#debugging-background-scripts)
+  - [Debugging options pages](#debugging-options-pages)
+  - [Debugging popups](#debugging-popups)
+  - [Debugging sidebars](#debugging-sidebars)
+  - [Debugging devtools panels](#debugging-developer-tools-pages-and-panels).
+
+Log messages from content scripts can be viewed in the developer tools of the tab where the content script runs instead of `about:debugging`, see:
+
+- [Debugging content scripts](#debugging-content-scripts)
+
```
rpl, line 111: (even just as a follow up) here or in that "Debugging content scripts" section we should add a warning box to mention explicitly [Bug 1410932](https://bugzilla.mozilla.org/show_bug.cgi?id=1410932) and to look for those errors in the Browser Console (after enabling the "Content Messages" toggle button). 

It may be better to do it in a separate follow up, we may want to double-check explicitly which are the cases that are still not logged in the webconsole, evilpie was the last one to give a try to that issue but from what I recall we were not able to get a reasonable solution that would cover all cases properly, but based on comment 43 it seems that [bug 1696756](https://bugzilla.mozilla.org/show_bug.cgi?id=1696756) may have covered the rejected promises.

src/content/documentation/develop/debugging.md
```diff
@@ -90,6 +92,41 @@ You can now drag the toolbox tab to a separate window, so you can place it along
...
+Log messages not associated with an extension script, such as the `stderr` output of the [nativeMessaging API](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/Native_messaging)
+can only be viewed via the Browser Console. This console contains messages from the whole browser, including your and other extensions.
+It occasionally contains more detailed information about errors reported to your extension. For more information, see:
```
rpl, line 114: It may be worth a mention that enabling that "Show Content Messages" toggle in the browser console may show logs not visible otherwise (not the case for the NativeMessaging.jsm though, and so we should try to not make it sound like it is mandatory for that).

rpl (approved):
looks good, follows some nits and some followups.

src/content/documentation/develop/debugging.md
Rob--W: This is documented at that page. I've kept this overview minimal to offer a good signal-to-noise ratio (I've added the index in the first place because it was not easy to find the relevant documentation unless you already knew what you were looking for).

src/content/documentation/develop/debugging.md
Rob--W: I've replaced the first sentence with `This console shows messages from the extension's background script; other logs may appear elsewhere.`.

rpl: @Rob--W that's not entire true, it shows the logs from all extension pages not just the background page.

Rob--W: Updated and added `and extension pages`

rpl: ty!

rpl (approved):

PR #1073, opened 2021-08-11
Update web-ext command reference

caitmuenster:
Two minor changes to web-ext command reference documentation: 

1) `--api-url-prefix` defaults to `https://addons.mozilla.org/api/v4` (was previously v3)
2) Remove reference to review in table explaining channels. The reference had set an expectation that self-distributed extensions are exempt from the review process.


muffinresearch (approved):
r+

PR #1131, opened 2021-10-16
Text remains visible during webfont load

jcabak:
- Text remains visible during webfont load (added `font-display: swap;`)
Now performance is much better
d


muffinresearch:
@jcabak thanks for the patch.

Issue #1139, opened 2021-11-01
Add-on Policy Changes 2021


kewisch:
We will be updating our policies, using this issue to track the documentation work.

PR #1140, opened 2021-11-01
Stage add-on policy changes (#1139)

kewisch:
Here are the changes for the new policy. I don't know if EW has a way to add an item with a sidebar, that isn't actually part of the sidebar, but I've added banners to both pages to switch back and forth.

### DIFF
--- a/src/content/documentation/publish/add-on-policies-dec-2021.md
+++ b/src/content/documentation/publish/add-on-policies-dec-2021.md
@@ -1,0 +1,254 @@
+---
+layout: sidebar
+title: Add-on Policies (December 2021)
+permalink: /documentation/publish/add-on-policies-dec-2021/
+topic: Publish
+tags: [add-ons, review-policy]
+contributors:
+  [
+    kewisch,
+    rebloor,
+    wagnerand,
+    atsay,
+    jvillalobos,
+    wbamberg,
+    kmaglione,
+  ]
+last_updated_by: kewisch
+date: 2021-11-03
+---
+
+<!-- Page Hero Banner -->
+
+{% capture page_hero_banner_content %}
+
+::: note
+This is a preview of the policies effective December 1st, 2021. You may wish to view our [current policies](../add-on-policies/) as well.
+:::
+
+# Add-on Policies
+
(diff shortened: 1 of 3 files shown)


muffinresearch:
> Here are the changes for the new policy. I don't know if EW has a way to add an item with a sidebar, that isn't actually part of the sidebar, but I've added banners to both pages to switch back and forth.

I'm not sure what you mean by this. Do you mean the TOC in the side-bar? To add that you'd need to define the subnavitems like the current policy does in the pages.json See https://github.com/mozilla/extension-workshop/blob/0ea84db0db16930acefc61655ec1a4cf70f04e8c/src/data/pages.json

src/content/documentation/publish/add-on-policies-dec-2021.md
```diff
@@ -0,0 +1,254 @@
...
+
+{% capture content_with_toc %}
+
+## No Surprises
+
+Users should be able to easily discern what the functionality of your add-on is and should not be presented with unexpected user experiences after installing it.  The add-on should have an easy-to-read description about everything it does, and any information it collects. Please consult our best practices guide for [creating an appealing listing](/documentation/develop/create-an-appealing-listing/).
+
+### Unexpected features
+
+“Unexpected” features are those that are unrelated to the add-on’s primary function, and are not intuitive from the add-on name or description to be expected by a user installing that add-on. This includes features that impact user privacy or security, make unexpected changes to web content, change default settings like the new tab page, homepage, or search engine, or are not related to the add-on’s core function(s).
```
muffinresearch, line 54: This sentence doesn't read well to me.

muffinresearch (approved):
r+wc 

Let me know if you need another review if you're updating the json data.

kewisch:
> > Here are the changes for the new policy. I don't know if EW has a way to add an item with a sidebar, that isn't actually part of the sidebar, but I've added banners to both pages to switch back and forth.
> 
> I'm not sure what you mean by this. Do you mean the TOC in the side-bar?
Sorry this was cryptic. I would like the new policies not to show in the TOC in the sidebar at all. They should only be linked from the banner at the top of the page. If I remove the entry from pages.json completely, this also means that the new policies page ends up with a blank sidebar.

When the policy goes live December 1st, I'll do another push to move the new policies to the original URL, at which point I'll also update the subpageitems and heading ids.

PR #1143, opened 2021-11-02
Final add-on policies 2021 (#1139)

kewisch:


muffinresearch (approved):
This will need re-basing but checking it locally it looks good and the TOC is working as expected.

Naturally it'll make sense to re-test this ahead of landing it closer to when it will need to be released. (Dec 1st)

muffinresearch (changes requested):
@kewisch I think this is missing deleting the staged distribution agreement file. 

Also does the effective date need to be updated in the distribution agreement (see line 48).

PR #1220, opened 2022-02-02
upgrade stylelint to 14.3.0, stylelint-config-standard to 24.0.0

tublitzed:
Upgrading stylelint and stylelint-config-standard.

A few related changes:

* Added newly required postcss-scss dep.
* Made _most_ of the suggested changes. A few small exceptions added related to selector-class-pattern, selector-id-pattern mostly to avoid a massive PR with changed selector/ids across a huge number of files. Filed https://github.com/mozilla/extension-workshop/issues/1219 to fix this in a separate PR.


src/assets/css/_main.scss
```diff
@@ -135,7 +137,7 @@ a:focus {
 
 .mzp-c-footer-links-social a:focus,
 .mzp-c-footer-primary-logo a:focus,
-.topNav a:focus,
+.top-nav a:focus,
```
tublitzed, line 140: I tried to avoid most of these (filed https://github.com/mozilla/extension-workshop/issues/1219) to come back and grab the rest. However, there were a few that had only a small number of views using them so figured I'd change the bits that didn't touch too many places.

Happy to revert for now though if folks would prefer we do this as a separate PR.

src/assets/css/_main.scss
```diff
@@ -1661,7 +1663,7 @@ a.tile-block-link {
     transform: rotate(360deg);
   }
 }
-
+/* stylelint-enable  selector-id-pattern */
```
tublitzed, line 1666: I hate doing this, but there's just too many files for it to make sense to roll in here. Filed https://github.com/mozilla/extension-workshop/issues/1219 to come back and address.

.stylelintrc
```diff
@@ -43,6 +45,12 @@
       ]
     }],
     "string-quotes": null,
+    "value-keyword-case": [
+      "lower",
+      {
+        "camelCaseSvgKeywords": true
```
tublitzed, line 51: See https://github.com/stylelint/stylelint/issues/5863

src/assets/css/_main.scss
```diff
@@ -1578,7 +1580,7 @@ a.tile-block-link {
   top: calc(100% + 1px);
   width: 100%;
   height: 1px;
-  background: rgba(0, 0, 0, 0.85);
+  background: rgba(0, 0, 0, 0.85%);
```
eviljeff, line 1583: `0.85` is different to `0.85%` isn't it?  (0.85 is 85%)

src/assets/css/_main.scss
```diff
@@ -1697,7 +1699,7 @@ a.tile-block-link {
 
     &:hover {
       color: $primary-color-hover;
-      box-shadow: #{rem-calc(0 2 8)} rgba(12, 12, 13, 0.33);
+      box-shadow: #{rem-calc(0 2 8)} rgba(12, 12, 13, 0.33%);
```
eviljeff, line 1702: same question about `%` here - and there are a few other places.

eviljeff (approved):
just questions about the `0.xx` to `0.xx%` changes

src/assets/css/_main.scss
tublitzed: Good catch :) Fix incoming for all of them.

eviljeff (approved):

PR #1247, opened 2022-02-28
Install unsigned extensions clarification

rebloor:
This address is the `dev-doc-needed` request on [Bug 1720923](https://bugzilla.mozilla.org/show_bug.cgi?id=1720923) Explicitly mention addon id is needed to install succssfully even in builds where "xpinstall.signatures.required" set to false is supported


rpl (approved):
lgtm, Thanks @rebloor.

PR #1252, opened 2022-03-02
Note about using site permission add-on in private browsing windows

rebloor:
As per [this Slack thread](https://mozilla.slack.com/archives/C02UVA681AP/p1646158101511189)


diox (approved):

src/content/documentation/publish/site-permission-add-on.md
```diff
@@ -99,6 +99,10 @@ If the user chooses to learn more, they are taken to the [Permission request mes
 
 Your service can now access the APIs.
 
+::: note
+To use your APIs in private browsing windows, the user needs to following the instructions to [enable extensions in private windows](https://support.mozilla.org/en-US/kb/extensions-private-browsing#w_enabling-or-disabling-extensions-in-private-windows).
```
mixedpuppy, line 103: To use the APIs enabled by your add-on in private browsing windows, the user needs to follow the instructions to

mixedpuppy (approved):
r+ after fixing "following".  I also found "your APIs" awkward so the comment offers a suggestion.

src/content/documentation/publish/site-permission-add-on.md
rebloor: I have fixed the "following" issue but this change may be ambiguous, it could be read that the add-on has enabled the APIs in the private window. I think the original text is clearer.

mixedpuppy: I was trying to address the "your APIs" which reads weird to me since it's not their APIs.  If this addon type was only webmidi, I would say "To use WebMIDI APIs in private browsing windows...", but it needs to be generic.

mixedpuppy (approved):
If "your APIs" can be changed somehow per my comment in the conversation tab, that would be nice.

src/content/documentation/publish/site-permission-add-on.md
rebloor: Got your point, it now reads "To use your service in…"

PR #1257, opened 2022-03-09
Avoid \d in bash regexps

### LINKED ISSUE #1236: Update circle config to support cherry-pick style tags
We are moving to weekly deploys for extension-workshop, so the tag format will be changing to `YYYY.MM.DD`, which works fine for tagging to stage, but if we need to do a cherry-pick style tag name (e.g., `2022.02.17-1-stage`) the current regex rejects that as a valid tag name. It would be good to be able to use that style of tag name where needed, so we should update https://github.com/mozilla/extension-workshop/blob/master/.circleci/config.yml to allow for tags with that format.

diox:
Fixes #1236

Hopefully for real this time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,7 +110,7 @@ jobs:
       - run:
           name: firefox extension workshop deployment
           command: |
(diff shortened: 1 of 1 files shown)


eviljeff (approved):
looks like we're good now up to the end of 2099 😬

Issue #1280, opened 2022-04-01
Clarify the methods for installing self hosted extensions


rebloor:
Starting from Firefox 100, a website's "install flow" has to meet the new "user activation" requirement Bug 1759737, in particular, see https://bugzilla.mozilla.org/show_bug.cgi?id=1759737#c12 and, for an example of a website that is hitting this new requirement, https://github.com/webcompat/web-bugs/issues/101685#issuecomment-1083039281 .

diox:
That's for `InstallTrigger` API though, which hasn't been (and should never be) documented on extensionworkshop - it's been deprecated for years, before being removed from MDN the documentation said:
>  This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

Websites should just link to the `.xpi`. Maybe https://extensionworkshop.com/documentation/publish/self-distribution/ could be clarified to say that.

rebloor:
@rpl thoughts?

rpl:
>> Websites should just link to the .xpi. Maybe https://extensionworkshop.com/documentation/publish/self-distribution/ could be clarified to say that.
>
> @rpl thoughts?

@rebloor That is basically the goal of this issue: to make sure we mention clearly from the docs included in or linked from https://extensionworkshop.com/documentation/publish/self-distribution/  what are the supported ways to trigger an addon installation from a third party website, even more important now that we need to prepare the extension developers to 3 changes to the addon install flows in Firefox >= 100:

- user activation requirement (required for both InstallTrigger and navigations to the XPI file) [Bug 1759737](https://bugzilla.mozilla.org/show_bug.cgi?id=1759737) is the one that some extension developers may more likely be hitting in the short run, and it is being enforced in Firefox >= 100 (I'm going to provide you some more details about this, in particular some example of how to make sure to trigger the installation through user activation)

- the additional cross-origin checks introduced in [Bug 1755950](https://bugzilla.mozilla.org/show_bug.cgi?id=1755950) are mainly making an existing requirement slightly more strict, most extension developers shouldn't be hitting it (and so we may just mention that there shouldn't be any third party frame between the top level page and the subframe that is going to be triggering the install flow)

[...]

PR #1292, opened 2022-04-21
Extra information regarding self-hosting web downloads of add-ons

rebloor:
Fixes https://github.com/mozilla/extension-workshop/issues/1280


src/content/documentation/publish/self-distribution.md
```diff
@@ -61,6 +61,52 @@ When you choose to distribute an add-on yourself, they can be installed using th
...
+{% capture content %}
+
+### Web download
+
+When distributing an add–on from your website, you have two options for making the signed add–on file available for download:
+
+- From a link on a webpage, like this:
+  ```html
+  <div id="example-option-1" class="install-ok">
+    <a href="https://addons.mozilla.org/firefox/downloads/file/3923300/facebook_container-2.3.2-fx.xpi">
```
rpl, line 73: @rebloor given that this page is related to self-distribution, I think it may be reasonable to use a fake xpi url that does point to a website that isn't addons.mozilla.org, e.g. "https://example.com/path/to/extension.xpi", wdyt?

rpl (commented):
Looks pretty good to me, in the review comments listed below just a small nit

src/content/documentation/publish/self-distribution.md
```diff
@@ -61,6 +61,52 @@ When you choose to distribute an add-on yourself, they can be installed using th
...
+      Install my add-on
+    </a>
+  </div>
+  ```
+  This is the recommended approach.
```
wagnerand, line 78: Do we want to move this up right below the chapter heading?

src/content/documentation/publish/self-distribution.md
rebloor: Done

src/content/documentation/publish/self-distribution.md
rebloor: That could make the presentation a bit awkward, I'd recommend leaving it here

src/content/documentation/publish/self-distribution.md
```diff
@@ -61,6 +61,52 @@ When you choose to distribute an add-on yourself, they can be installed using th
...
+    <script>
+      document.querySelector("#example-option-2 > button").onclick = () => {
+        window.location = "https://example.com/path/to/extension.xpi";
+      };
+    </script>
+  </div>
+  ```
+  If you take this approach, make sure that:
+  - The browser detects your JavaScript code as handling user input, for example, it is called from inside a DOM click event.
+  - If the installation is not triggered from the top-level frame of the website, none of the frames between the top-level page and the frame triggering the installation are a third party to the top-level page.
```
rpl, line 94: Not having any cross origin frames in between is actually a requirement also for installations triggered without any website JS code involved (and so also for a plain link to the xpi url).

rebloor: Updated, please check I've understood your feedback correctly

rpl: @rebloor thanks, :ok_hand:  yes the last change seems to be handling what I was pointing out, just one question:
- while looking to the interdiff I did notice that the same restriction is described with a slightly different wording in the two described approaches, any reason to don't use the exact same wording? 
(and eventually move that outside of the two approaches and into a note box as for the part that mention the deprecation of the non-standard InstallTrigger Web API).

src/content/documentation/publish/self-distribution.md
```diff
@@ -61,6 +61,53 @@ When you choose to distribute an add-on yourself, they can be installed using th
...
+  ```
+  If you take this approach, make sure that the browser detects your JavaScript code as handling user input, for example, it is called from inside a DOM click event.
+  
+When using either method, if the download link or triggering JavaScript is not on the top-level frame of the website, none of the frames between the top-level page and the frame containing the download or JavaScript can be third-party or cross-origin frames.
+
+::: note
+You may find references to using the `InstallTrigger` API to trigger a download from JavaScript. This API was never officially supported and will be removed from Firefox 102. Use is therefore not recommended.
+:::
+
+Using either method, when the user downloads the signed add-on file Firefox installs it.
```
rpl, line 101: There is another "using either method" a few lines before (at line 95), not a big deal but maybe we could use a slightly different wording in one of them.

rpl (approved):
looks good to me from a technical perspective, I'll leave to @wagnerand the final sign off.

rpl:
@wagnerand would you mind to take another look to this PR and sign it off if there isn't any remaining blocking issues from your perspective?

@rebloor (or @wagnerand) github seems to be suggesting to rebase this PR on a more recent main branch, but it doesn't seem it is detecting any conflicts. Would you mind to rebase this PR to prepare it to be merged as soon as it is signed off also by @wagnerand?

src/content/documentation/publish/self-distribution.md
wagnerand: Can we make it more prominent, that this is the recommended approach? It doesn't seem very visible as standard text and it being at the end of the section, in the middle of a larger chapter.

rebloor: I understand this is a fairly soft recommendation and therefore doesn't particularly need to be highlighted @rpl any thoughts?

src/content/documentation/publish/self-distribution.md
rebloor: Good point @rpl as in the first instance we mention both methods in the sentence, I've removed this introductory phrase from there as it seems somewhat unnecessary.

src/content/documentation/publish/self-distribution.md
willdurand: I moved this sentence before the code snippet because the previous version made this sentence rather invisible.

willdurand:
about the last commit, got a r+ from @rpl on slack and from @wagnerand on matrix, 🚢

PR #1298, opened 2022-04-29
MV3 Background event pages updates

rebloor:
- Modified the introduction to debugging background pages to mention persistent and non-persistent scripts
- Added notes about non-persistent scripts remaining open while toolbox is open and the ability to close background pages from about:debugging


src/content/documentation/develop/debugging.md
```diff
@@ -145,6 +147,8 @@ To get started, open your background script in **Sources**.
 
 As you run your extension, the **Console** displays logged output, including calls to [`console.log()`](https://developer.mozilla.org/docs/Web/API/Console/log) made by the extension's background scripts and any errors the browser raises as it executes these scripts. Note that the console shows all errors raised by the browser, not just errors related to your extension's code.
 
+When you debug a non-persistent background script, the background script won't go idle while the toolbox is open. However, if you need to terminate the background page, you can do so in `about:debugging`.
```
mixedpuppy, line 150: IIUC dev tools also has a button.  @rpl

mixedpuppy (changes requested):
Lets verify my comment with @rpl first.

src/content/documentation/develop/debugging.md
rpl: @mixedpuppy The "Terminate background script" button is available in the about:debugging addon card and so the way line 150 describes where to find it looks correct (the toolbox is the actual set of tools opened and attached to the extension in a new tab after the extension developer has clicked the "Inspect" button).

We may use `DevTools Toolbox` to refer to the toolbox if we want to better match how the Firefox DevTools team usually refers to it.

rebloor: Thanks @rpl I'll create an issue for the use of `DesTools Toolbox` as it affects the entire article. I assume the "Terminate background script" button only appears for a non-persistent background page - when I have finished the core changes for the developer preview I'll go back and convert the [notify-link-clicks-i18n](https://github.com/mdn/webextensions-examples/tree/master/notify-link-clicks-i18n) example and add a screenshot.

mixedpuppy (approved):

src/content/documentation/develop/debugging.md
```diff
@@ -135,7 +135,9 @@ We use the [notify-link-clicks-i18n](https://github.com/mdn/webextensions-exampl
...
+[Background scripts](https://developer.mozilla.org/Add-ons/WebExtensions/Background_scripts) enable an extension to monitor and react to events in the browser, such as navigating to a new page, removing a bookmark, or closing a tab. These scripts can be persistent or non-persistent. Persistent background scripts remain loaded for the lifetime of the extension. In contrast, non-persistent background scripts load when needed to respond to an event and unloaded when they become idle. Non-persistent background scripts are recommended because they consume less of the browser's resources.
```
rpl, line 138: `Non-persistent background scripts are recommended because they consume less of the browser's resources.` I'm not sure this page is the right place where to provide guidance about when a persistent vs. a non-persistent background page is required, or advised, without risking to have to repeat details that may be provided in another place where they fit better (e.g. in the API doc for the "background" manifest property).

e.g. we don't call out here that in MV3 only non-persistent background scripts are allowed, not that in MV2 the "persistent" property is still ignored.

I would propose to but out the last phrase (just the "Non-persistent background scripts are recommended ....").

rpl (approved):
I think that even with my approval, this PR would still be blocked on the CI job successfully run on the PR, which may actually not happen without the author re-authorizing Circle CI to allow it to access the github repos part of the mozilla organization (which is a mandatory requirement after we migrated to Github SAML).

See https://support.circleci.com/hc/en-us/articles/360043002793-Troubleshooting-CircleCI-access-after-enabling-Github-SSO, I had to do the same on Friday to make sure Circle CI was going to run a job for an addons-linter PR that I created after the Github SAML migration).

@rebloor would you mind to give a read to the Circle CI troubleshooting article linked above and go through those steps? I may take care of this one by recreating it again with my github user, but it would be needed for any other PR you create for repos in the mozilla organization and so it would be good to confirm that the circle ci jobs are going to run in the pull request you are going to create from now on (and other ones that may be already created and pending on a CI job to run successfully to be mergeable).

Issue #1315, opened 2022-05-16
Anchor elements with code inside are not readable in current style on extension workshop site


willdurand:
This issue was initially reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1768377 but it belongs in this repo.

---

## Steps to reproduce:

1. Open any page under the domain: https://extensionworkshop.com/documentation/develop - that contains a code element inside an anchor element.
    - An example page that has this issue is: https://extensionworkshop.com/documentation/develop/extensions-and-the-add-on-id/
2. On this page, scroll down to the section: When do you need an add-on ID?
    - The issue can be seen all over the page, but this particular section has a list of code elements that are also external links.

## Actual results:

Any text contained in `<a><code>` is hard to read because "code" elements are colored blue - #015fdf, while anchor `a` elements have a dark blue background color - #20123a, resulting in blue text over dark blue background. The contrast is negligible, so the text is really hard to read.

## Expected results:

There are two solutions:

- Either the code (only if inside an anchor element) should have white (or some shade of white) colored text, with the default dark blue background color of anchor element.
- Or, the anchor element (only if some code element is a child) should have no background color, i.e., transparent (or some color that has a high contrast to blue).

willdurand:
Looks like this is a regression because it used to be better, see: https://github.com/mozilla/extension-workshop/pull/746

PR #1317, opened 2022-05-17
Fix inline code style

willdurand:
Fixes https://github.com/mozilla/extension-workshop/issues/1315

---

This is essentially re-applying #746 because Foundation has [changed *something*](https://github.com/foundation/foundation-sites/compare/v6.7.3...v6.7.4#diff-fd8a074f3acce03326b1dc2d6e457abebd1ca16673b4ff0438c36779cd346fcc).

## Screenshots

Before:

Screenshot 2022-05-17 at 15 49 22

After:

Screenshot 2022-05-17 at 15 49 11

### DIFF
--- a/src/assets/css/_main.scss
+++ b/src/assets/css/_main.scss
@@ -254,10 +254,11 @@ table {
 // ------
 
 code {
-  border-radius: $global-radius;
-  overflow-wrap: break-word;
   background-color: $light-gray;
+  border-radius: $global-radius;
   color: $black;
+  overflow-wrap: break-word;
+  padding: rem-calc(1 5);
 
   &.nobackground {
     background: none;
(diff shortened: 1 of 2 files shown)


src/assets/css/_main.scss
```diff
@@ -254,10 +254,11 @@ table {
 // ------
 
 code {
-  border-radius: $global-radius;
-  overflow-wrap: break-word;
   background-color: $light-gray;
+  border-radius: $global-radius;
   color: $black;
+  overflow-wrap: break-word;
+  padding: rem-calc(1 5);
```
willdurand, line 261: I added `padding` and re-sorted the rules because that's what we do in addons-frontend :p

eviljeff:
I'm guessing this was https://github.com/mozilla/extension-workshop/pull/1142 that renovate auto-merged, but that was 6 months ago - has it been broken since last year and we just never noticed?  (which is fine, I'm only wondering)

willdurand:
> I'm guessing this was #1142 that renovate auto-merged, but that was 6 months ago - has it been broken since last year and we just never noticed? (which is fine, I'm only wondering)

Yeah, I think it isn't new. Rob said he noticed that as well but we received a bug in Bugzilla recently, which is why I filed the issue. I am writing some new docs and this was annoying so I thought I'd fix it.

eviljeff (approved):

Issue #1310, opened 2022-05-11
Update "Extensions and the add-on ID" page to mention the MV3+ requirement


willdurand:
We are going to require the extension ID for MV3 add-ons, see: https://github.com/mozilla/addons-linter/issues/4303. We need to update [this page](https://extensionworkshop.com/documentation/develop/extensions-and-the-add-on-id/) accordingly.

willdurand:
cc @wagnerand

PR #1318, opened 2022-05-17
Add some notes about the add-on ID being required for MV3 extensions

willdurand:
Fixes https://github.com/mozilla/extension-workshop/issues/1310

---

We can probably improve a bit in the future but at least we have the
basics covered. I removed the mentions to SDK/XUL and FF 48 because...
this is super old IMO.


wagnerand (approved):

willdurand:
(rebase)

rpl (approved):

PR #1339, opened 2022-06-17
document `--firefox-preview` option for `web-ext run`

willdurand:


src/content/documentation/develop/web-ext-command-reference.md
```diff
@@ -239,6 +240,14 @@ web-ext run --target=firefox-android --firefox-apk=org.mozilla.firefox
 
 Environment variable: `$WEB_EXT_FIREFOX_APK`
 
+#### `--firefox-preview`
+  
+Turn on developer preview features in Firefox. This option accepts multiple values, although it currently only supports the `mv3` value, which is also the default value.
+
```
rpl, line 246: this may also be a good place to mention what the value means, given that we only have one maybe we could just inline something short right here? wdyt?

Nothing too long, I was thinking to just something like: 
> ..., which is also the default value and can be used by developers that are testing out their extensions with Firefox Manifest Version 3 support (without having to manually flipping the related preferences, see [MV3 migration guide](https://extensionworkshop.com/documentation/develop/manifest-v3-migration-guide/))

and eventually if we will ever end up have multiple values supported we can change the format to a unordered list.

willdurand: ah yeah, it makes sense to document the values...

rpl (approved):

PR #1364, opened 2022-07-24
Remove legacy extensions and other outdated content

rebloor:
Removing various outdated content, including:

- legacy extensions
- fennec development

Dependent on [Redirects for outdated page #1368](https://github.com/mozilla/extension-workshop/issues/1368) to add redirects before merging.


rebloor:
@willdurand it is looking unlikely that we can get redirects for the removed pages. I've had no response for the best part of six months. If you're OK, let's go ahead and remove these, and if we get the redirects, we get them one day

rebloor:
@willdurand now that we have the redirects in place, would you like to go ahead and merge this one?

PR #1372, opened 2022-08-03
Document required content type for add-on web download install flow

mryanmurphy:
I was confused why my self hosted add-on did not have a prompt to install when downloading. Documenting my fix here for those who were using the deprecated `InstallTrigger` method.


willdurand:
(I think this is fine but just needs confirmation from @rpl)

src/content/documentation/publish/self-distribution.md
```diff
@@ -50,7 +50,7 @@ For more information on how to submit an add-on for distribution on AMO or self-
 
 When you choose to distribute an add-on yourself, they can be installed using the following methods:
 
-- [Web download](#web-download)—make your extension available on a suitable web accessible server and when the user downloads the signed add-on file Firefox installs it.
+- [Web download](#web-download)—make your extension available on a suitable web accessible server and when the user downloads the signed add-on file Firefox installs it. The signed add-on file must have a `Content-Type: application/x-xpinstall` HTTP response header.
```
rpl, line 53: Nit: The web accessible server must serve the signed addon xpi file with `Content-Type: application/x-xpinstall` in the HTTP response headers.

rpl (approved):
This is definitely the case, thanks @mryanmurphy for updating the docs to explicitly point that out.

In the inline comment below I'm suggesting a slightly reworded version, but it is just a nit and so I'm also approving this PR as is.

src/content/documentation/publish/self-distribution.md
mryanmurphy: Updated using reworded suggestion

mryanmurphy:
Thanks @rpl, updated wording based on your suggestion.

willdurand (approved):

PR #1408, opened 2022-09-23
Remove LaserCat as an example for good onboard page

Hamatti:
Since Laser Cat is no longer available in addons.mozilla.org, it needs to be removed from the examples.

Also changes the prepending paragraph from plural to singular as we only now have one example.


src/content/documentation/develop/onboard-upboard-offboard-users.md
```diff
@@ -126,27 +126,6 @@ Here are some examples of good onboarding pages:
 
 <!-- Two Column Body Module -->
 
-<section class="module">
-<article class="module-content grid-x grid-padding-x align-middle">
-<div class="cell small-12 medium-6 medium-order-2">
-
-![Example onboarding page that details how to use the extension's features and turn them on and off](/assets/img/documentation/develop/Laser_cat_onboarding.png)
```
willdurand, line 133: Can we delete the image in the repo too?

Hamatti: Good catch, I removed it.

Issue #1427, opened 2022-10-14
Document simpler version string format


willdurand:
We should add an entry about the new version string format in the MV3 migration guide. See: https://bugzilla.mozilla.org/show_bug.cgi?id=1793925

PR #1430, opened 2022-10-17
Add a note about the version format for MV3

willdurand:
Fixes https://github.com/mozilla/extension-workshop/issues/1427

### DIFF
--- a/src/content/documentation/develop/manifest-v3-migration-guide.md
+++ b/src/content/documentation/develop/manifest-v3-migration-guide.md
@@ -5,8 +5,8 @@ permalink: /documentation/develop/manifest-v3-migration-guide/
 topic: Develop
 tags: [webextensions, api, firefox]
 contributors: [rebloor, willdurand, erosman, Klestofer]
-last_updated_by: Klestofer
-date: 2022-09-12
+last_updated_by: willdurand
+date: 2022-10-17
 ---
 
 <!-- Page Hero Banner -->
@@ -283,13 +283,14 @@ In Manifest v2, Firefox extensions support the use of the `chrome.*` namespace w
 ## Migration checklist
 
 - Update the manifest.json key `manifest_version` to `3`.
-- If your extension adds a search engine, add a local icon and reference it in the  manifest.json key `chrome_settings_overrides.search_provider.favicon_url`.
+- If your extension adds a search engine, add a local icon and reference it in the manifest.json key `chrome_settings_overrides.search_provider.favicon_url`.
 - Remove any host permissions from the manifest.json keys permissions and `optional_permissions` and add them to the `host_permissions` key.
 - Rename the manifest.json key `browser_action` to `action` and update any API references from `browser.browserAction` to `browser.action`.
 - Convert background pages to be non-persistent.
 - Move the extension’s CSP to the the manifest.json key ‘content_security_policy.extension_pages’ and update the CSP to conform to Manifest V3 requirements.
 - Move any arbitrary strings executed as scripts to files and update your code to use the Scripting API.
 - The add-on ID is required to publish your extension. Make sure to add one in the `browser_specific_settings.gecko.id` manifest.json key.
+- The format of the `version` manifest.json key has evolved and became simpler: the `version` must be a string with 1 to 4 numbers separated with dots (e.g. `1.2.3.4`). Each number can have up to 9 digits and leading zeros are not allowed (e.g. `2.01` is forbidden but `2.1` is allowed).
 
 {% endcapture %}
 {% include modules/one-column.liquid


src/content/documentation/develop/manifest-v3-migration-guide.md
```diff
@@ -283,13 +283,14 @@ In Manifest v2, Firefox extensions support the use of the `chrome.*` namespace w
...
+- If your extension adds a search engine, add a local icon and reference it in the manifest.json key `chrome_settings_overrides.search_provider.favicon_url`.
 - Remove any host permissions from the manifest.json keys permissions and `optional_permissions` and add them to the `host_permissions` key.
 - Rename the manifest.json key `browser_action` to `action` and update any API references from `browser.browserAction` to `browser.action`.
 - Convert background pages to be non-persistent.
 - Move the extension’s CSP to the the manifest.json key ‘content_security_policy.extension_pages’ and update the CSP to conform to Manifest V3 requirements.
 - Move any arbitrary strings executed as scripts to files and update your code to use the Scripting API.
 - The add-on ID is required to publish your extension. Make sure to add one in the `browser_specific_settings.gecko.id` manifest.json key.
+- The format of the `version` manifest.json key has evolved and became simpler: the `version` must be a string with 1 to 4 numbers separated with dots (e.g. `1.2.3.4`). Each number can have up to 9 digits and leading zeros are not allowed (e.g. `2.01` is forbidden but `2.1` is allowed).
```
Rob--W, line 293: Leading zeroes were already invalid before in the linter, so it's unlikely to help anyone.

I suggest to simplify the description to only what's been changed and remove the "and leading... allowed)". To highlight the change, put "Letters are no longer allowed." after "simpler:". Then readers won't have to wonder what changed and what they have to do 

MDN should list the constraints in the documentation, and the EW can link to that article.

Rob--W (changes requested):

src/content/documentation/develop/manifest-v3-migration-guide.md
willdurand: > Leading zeroes were already invalid before in the linter, so it's unlikely to help anyone.

`1.01a` was accepted by the linter.

Rob--W: But there is a letter in there which triggered the special case.

The "leading zero" remark could raise questions about the validity of a plain zero.

I'm trying to find a balance between what people have to read and what actionable parts there are. The sniff test "does version contain a letter" covers most if not all cases. For a precise description a reference to specific description seems appropriate and sufficient.

Something like this would work:

> "The `version` key in manifest.json has a stricter format. Letters are no longer allowed, only up to four numbers separated with a dot. For the precise description, see the [version documentation on MDN]"

willdurand: I made some edits. That should be good until we have the page on MDN updated (without that being a blocker).

src/content/documentation/develop/manifest-v3-migration-guide.md
```diff
@@ -283,13 +283,14 @@ In Manifest v2, Firefox extensions support the use of the `chrome.*` namespace w
...
 - Remove any host permissions from the manifest.json keys permissions and `optional_permissions` and add them to the `host_permissions` key.
 - Rename the manifest.json key `browser_action` to `action` and update any API references from `browser.browserAction` to `browser.action`.
 - Convert background pages to be non-persistent.
 - Move the extension’s CSP to the the manifest.json key ‘content_security_policy.extension_pages’ and update the CSP to conform to Manifest V3 requirements.
 - Move any arbitrary strings executed as scripts to files and update your code to use the Scripting API.
 - The add-on ID is required to publish your extension. Make sure to add one in the `browser_specific_settings.gecko.id` manifest.json key.
+- The format of the top-level `version` manifest.json key has evolved and became simpler: letters and other previously allowed symbols are no longer accepted. The value must be a string with 1 to 4 numbers separated with dots (e.g. `1.2.3.4`). Each number can have up to 9 digits and leading zeros are not allowed (e.g. `2.01` is forbidden but `2.1` is allowed).
```
Rob--W, line 293: This sentence is thrice as long as the other bullet points. This is a checklist and the points should be brief.

This level of detail is perfect for a changelog and/or reference documentation. Could you move it to https://github.com/mdn/content/blob/main/files/en-us/mozilla/firefox/releases/107/index.md and tag Richard for content review? Or just open an issue in mdn/content, assigned to him with the above text as an example. Along with the changelog, the following needs to be updated (I'm willing to take over the review etc if you'd like):

- https://github.com/mdn/content/blob/main/files/en-us/mozilla/add-ons/webextensions/manifest.json/version/format/index.md
- https://github.com/mdn/content/blob/main/files/en-us/mozilla/add-ons/webextensions/manifest.json/version/index.md

This is the kind of brevity that I was thinking of:

```
- Ensure that the top-level manifest.json `version` key is a string consisting of digits separated with up to 3 dots. Letters are no longer allowed. For details, see [version format](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/version/format).
```

renders as

> - Ensure that the top-level manifest.json `version` key is a string consisting of digits separated with up to 3 dots. Letters are no longer allowed. For details, see [version format](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/version/format).

Maybe even without "Letters are no longer allowed." if you think that the suggestion is too long.

willdurand: I understand what you're saying here but the MDN are (or will be) outdated so linking to them now doesn't seem to be a great idea. As you pointed out, it isn't just about adding some release notes, there are also two pages to update (one requiring significant changes).

I can add your suggestion and move the more verbose paragraph in a section above.

willdurand: > move the more verbose paragraph in a section above.

We can consider removing it later if we want to but I want this change to be documented somewhere since the linter and FF will soon show warnings.

willdurand: > This level of detail is perfect for a changelog and/or reference documentation. Could you move it to [mdn/content@`main`/files/en-us/mozilla/firefox/releases/107/index.md](https://github.com/mdn/content/blob/main/files/en-us/mozilla/firefox/releases/107/index.md?rgh-link-date=2022-10-17T22%3A25%3A27Z) and tag Richard for content review? Or just open an issue in mdn/content, assigned to him with the above text as an example. Along with the changelog, the following needs to be updated (I'm willing to take over the review etc if you'd like):
> 
>     * [https://github.com/mdn/content/blob/main/files/en-us/mozilla/add-ons/webextensions/manifest.json/version/format/index.md](https://github.com/mdn/content/blob/main/files/en-us/mozilla/add-ons/webextensions/manifest.json/version/format/index.md?rgh-link-date=2022-10-17T22%3A25%3A27Z)
> 
>     * [https://github.com/mdn/content/blob/main/files/en-us/mozilla/add-ons/webextensions/manifest.json/version/index.md](https://github.com/mdn/content/blob/main/files/en-us/mozilla/add-ons/webextensions/manifest.json/version/index.md?rgh-link-date=2022-10-17T22%3A25%3A27Z)

108, no? Also, there is already a dev-doc-needed tag in the bug so the release notes and the two other pages linked above should be updated for that bug. I'll send the link to this PR to Richard.

Rob--W: Yes 108 now.

Rob--W (approved):

PR #1449, opened 2022-11-04
Fix MDN manifest_version link in MV3 migration guide

advaith1:
Currently this links to the page about the unrelated `version` field, rather than the `manifest_version` field.


diox:
Thanks!

PR #1452, opened 2022-11-09
Add mention of --chromium-binary to --target documentation

Hamatti:
When using `--target chromium`, the user can further specify which Chromium browser they want to use with `--chromium-binary`.

This is not clear from current documentation and comes up rather often so decided to document here.

Also, to maintain consistency, added link to `--android-device` one row above now that we support anchor links in the docs.

I was pondering between two options:

1) Current option that adds a mention to the table but doesn't specifically call out that it can be used for other browsers than Chrome

2) Instead of current change, add a mention below the table:

>If target is `chromium`, you can specify which binary to run with `--chromium-binary` to run other Chromium based browsers than Chrome.

I'm open for opinions on that.


Hamatti:
This helps in cases like https://github.com/mozilla/web-ext/issues/2516

willdurand:
(I am surprised that prettier or another tool does not say anything about the format of the table but 🤷)

Issue #1461, opened 2022-11-16
Sidebar contents is missing


diox:
STR:
- Visit any page on extension-workshop with some sidebar contents, like /documentation/develop/

Expected results:
- Sidebar contents is shown and links work

Actual results:
- Sidebar appear to be empty

PR #1462, opened 2022-11-16
Revert "selector-not-notation" changes - they are breaking the styles

diox:
Fixes https://github.com/mozilla/extension-workshop/issues/1461


Hamatti (approved):
Tested on my local environment, can confirm that it works.

PR #1563, opened 2023-03-09
Fix 1368

bqbn:


willdurand:
@diox could you please take a look?

PR #1564, opened 2023-03-10
Fixes event pages and backward compatibility section

rebloor:
Resolves formatting issues in the event pages and backward compatibility section. Also, add an anchor for the section.


Rob--W (approved):
Thanks for adding the header. Please verify that the header exists after merging/publishing.

PR #1596, opened 2023-04-22
fix links to amo signing api docs

djbrown:
which was removed from v5
and is only available on v4 (frozen/stable)
see also https://github.com/mozilla/addons-server/issues/20560


willdurand:
Hey @eviljeff, could you please review this PR? Thanks

eviljeff (approved):
r+
We'll have to make sure to update these docs when `--use-submission-api` becomes the default.

PR #1670, opened 2023-08-05
Fix strange link/typo

rugk:
I assume this was a mistake and this should not be rendered as that:
> Then pass the permissions you want granted to WebExtAPIRef("permissions.request"), which prompts the user to grant the permissions


rebloor (approved):

rebloor:
Thanks @rugk

PR #1698, opened 2023-09-18
Add office hours resources

dotproto:
Add links to office hours sessions for developers targeting Firefox for Android.


src/content/community.md
```diff
@@ -52,6 +53,39 @@ There are several ways to stay up-to-date with Firefox add-on technology and con
 
 Stay up to date with the latest information about add-on technology on the [Add-ons blog](https://blog.mozilla.org/addons).
 
+### Office hours <span id="office-hours"></span>
+
+The Firefox Add-ons team holds one-on-one office hours sessions to help developers make their extensions available on Android. Developres can also use the [Firefox for Android Add-ons](https://discourse.mozilla.org/c/add-ons/android/9393) forum to discuss issues asynchronously.
```
rpl, line 58: Typo: Developres => Developers

rpl (commented):
I haven't tried it locally yet but I noticed a typo, see inline review comment below.

src/content/community.md
dotproto: Fixed and force-pushed the update.

Rob--W:
I'd like to take a closer look at the need to load a third-party script here.

dotproto:
The third party script isn't strictly required.

The primary reason I included it was that I wanted https://extensionworkshop.com/community/#office-hours to be the primary URL that developers use to book time rather than the [Google Calendar URL](https://calendar.google.com/calendar/appointments/schedules/AcZssZ3yNnoZNBdJCTiJjqziXs9GoCYpfW9XZFkB0QWKNy9dfoqqPlCw7UsYh48Ff0i3lZcA5YzygwrU). I was worried that just linking to Google Calendar might encourage users to bookmark that page. 

I was originally planning to embed the booking page in an iframe. To quickly revisit this idea, we could hide the iframe  in a details element. Alternatively, we could dynamically inject an iframe when the user clicks a "Book a session" link/button, which is essentially what the 3rd party script is doing. Finally, we could create a separate dedicated page for booking a session and have the calendar statically sandboxed in an iframe on that page. For example, https://extensionworkshop.com/community/office-hours/android-addons.

src/content/community.md
```diff
@@ -52,6 +53,39 @@ There are several ways to stay up-to-date with Firefox add-on technology and con
...
+/*
+ * Google's scripts inject a button styled with "Roboto" but doesn't load that
+ * font. Unfortunately, that means the font-family falls back to Times in
+ * Firefox and Chrome. Since we can't set an inline style on this element, the
+ * next best thing we can do is force the style on buttons in this this section.
+ */
+#connect-with-the-community button {
+  font-family: inherit !important;
+}
+</style>
```
willdurand, line 87: It would be good to move this to a scss file:

```diff
diff --git a/src/assets/css/_hacks.scss b/src/assets/css/_hacks.scss
new file mode 100644
index 0000000..423aca7
--- /dev/null
+++ b/src/assets/css/_hacks.scss
@@ -0,0 +1,9 @@
+/**
+ * Google's scripts inject a button styled with "Roboto" but doesn't load that
+ * font. Unfortunately, that means the font-family falls back to Times in
+ * Firefox and Chrome. Since we can't set an inline style on this element, the
+ * next best thing we can do is force the style on buttons in this this section.
+ */
+#connect-with-the-community button {
+  font-family: inherit;
+}
diff --git a/src/assets/css/styles.scss b/src/assets/css/styles.scss
index b3735ec..390d6bf 100644
--- a/src/assets/css/styles.scss
+++ b/src/assets/css/styles.scss
@@ -20,6 +20,7 @@ CSS file.
 @import 'contentguidelines';
 @import 'fullwidth';
 @import 'sidebar';
+@import 'hacks';
 
 // Select the components you want to use
 @include foundation-global-styles;
diff --git a/src/content/community.md b/src/content/community.md
index 5839c8a..3c99eae 100644
--- a/src/content/community.md
+++ b/src/content/community.md
@@ -66,7 +66,7 @@ The Firefox Add-ons team holds one-on-one office hours sessions to help develope
   window.addEventListener('load', function() {
     calendar.schedulingButton.load({
       url: 'https://calendar.google.com/calendar/appointments/schedules/AcZssZ3yNnoZNBdJCTiJjqziXs9GoCYpfW9XZFkB0QWKNy9dfoqqPlCw7UsYh48Ff0i3lZcA5YzygwrU?gv=true',
-      color: '#039BE5',
+      color: '#015fdf',
       label: 'Book an appointment',
       target,
     });
@@ -74,17 +74,6 @@ The Firefox Add-ons team holds one-on-one office hours sessions to help develope
 })();
 </script>
 <!-- end Google Calendar Appointment Scheduling -->
-<style>
-/*
- * Google's scripts inject a button styled with "Roboto" but doesn't load that
- * font. Unfortunately, that means the font-family falls back to Times in
- * Firefox and Chrome. Since we can't set an inline style on this element, the
- * next best thing we can do is force the style on buttons in this this section.
- */
-#connect-with-the-community button {
-  font-family: inherit !important;
-}
-</style>
 
 ### Questions about add-on development
 
```

Also the color of the button should probably be like the other one at the top.

src/content/community.md
```diff
@@ -52,6 +53,39 @@ There are several ways to stay up-to-date with Firefox add-on technology and con
 
 Stay up to date with the latest information about add-on technology on the [Add-ons blog](https://blog.mozilla.org/addons).
 
+### Office hours <span id="office-hours"></span>
+
+The Firefox Add-ons team holds one-on-one office hours sessions to help developers make their extensions available on Android. Developers can also use the [Firefox for Android Add-ons](https://discourse.mozilla.org/c/add-ons/android/9393) forum to discuss issues asynchronously.
+
+<!-- Google Calendar Appointment Scheduling begin -->
+<link href="https://calendar.google.com/calendar/scheduling-button-script.css" rel="stylesheet">
+<script src="https://calendar.google.com/calendar/scheduling-button-script.js" async></script>
```
willdurand, line 62: I don't think this will work once deployed because `https://calendar.google.com/calendar/` isn't specified in the CSP.

Rob--W:
> The third party script isn't strictly required.
> 
> The primary reason I included it was that I wanted https://extensionworkshop.com/community/#office-hours to be the primary URL that developers use to book time rather than the [Google Calendar URL](https://calendar.google.com/calendar/appointments/schedules/AcZssZ3yNnoZNBdJCTiJjqziXs9GoCYpfW9XZFkB0QWKNy9dfoqqPlCw7UsYh48Ff0i3lZcA5YzygwrU). I wanted the flexibility to change the underlying service we use to schedule office hours and was worried that just linking to Google Calendar might encourage users to bookmark or share that page.
> 
> Before I saw that Google Calendar provides the snippet I used, I was originally planning to embed the booking page in an iframe. To quickly revisit this idea, we could hide the iframe in a details element. Alternatively, we could dynamically inject an iframe when the user clicks a "Book a session" link/button, which is essentially what the 3rd party script is doing. Finally, we could create a separate dedicated page for booking a session and have the calendar statically sandboxed in an iframe on that page. For example, https://extensionworkshop.com/community/office-hours/android-addons.

If the result is still acceptable, I would very much prefer an iframe over an external script.

In either case, you have to allowlist the frame (and/or script) destination in the CSP. The CSP is configured at https://github.com/mozilla/extension-workshop/blob/2e5468e85375dd9e82ebac61a78271cee3f9fc8e/.utils/deploy.sh#L37-L47

src/content/community.md
dotproto: Good call. I'll revise the PR with this and [Rob's comment](https://github.com/mozilla/extension-workshop/pull/1698#issuecomment-1726184721) in mind.

README.md
```diff
@@ -10,7 +10,7 @@ If you would like to update content or other resources on Firefox Extension Work
 
 ## Development Guide: Getting Started
 
-These instructions will get you a copy of the project up and running on your local machine for development and testing purposes. 
+These instructions will get you a copy of the project up and running on your local machine for development and testing purposes.
```
rebloor, line 13:
```suggestion
These instructions get a copy of the project up and running on your local machine for development and testing purposes.
```

src/content/community.md
```diff
@@ -52,6 +53,39 @@ There are several ways to stay up-to-date with Firefox add-on technology and con
 
 Stay up to date with the latest information about add-on technology on the [Add-ons blog](https://blog.mozilla.org/addons).
 
+### Office hours <span id="office-hours"></span>
+
+The Firefox Add-ons team holds one-on-one office hours sessions to help developers make their extensions available on Android. Developers can also use the [Firefox for Android Add-ons](https://discourse.mozilla.org/c/add-ons/android/9393) forum to discuss issues asynchronously.
```
rebloor, line 58: Make this more personal:
```suggestion
The Firefox Add-ons team holds one-on-one office hours sessions to help you make your extensions available on Android. This is in addition to the [Firefox for Android Add-ons](https://discourse.mozilla.org/c/add-ons/android/9393) forum, where you can seek advice and assistance anytime.
```

src/content/community.md
```diff
@@ -52,6 +53,39 @@ There are several ways to stay up-to-date with Firefox add-on technology and con
...
+<link href="https://calendar.google.com/calendar/scheduling-button-script.css" rel="stylesheet">
+<script src="https://calendar.google.com/calendar/scheduling-button-script.js" async></script>
+<script>
+(function() {
+  var target = document.currentScript;
+  window.addEventListener('load', function() {
+    calendar.schedulingButton.load({
+      url: 'https://calendar.google.com/calendar/appointments/schedules/AcZssZ3yNnoZNBdJCTiJjqziXs9GoCYpfW9XZFkB0QWKNy9dfoqqPlCw7UsYh48Ff0i3lZcA5YzygwrU?gv=true',
+      color: '#039BE5',
+      label: 'Book an appointment',
```
rebloor, line 70: To align with earlier wording
```suggestion
      label: 'Book a session',
```

src/content/community.md
```diff
@@ -52,6 +53,39 @@ There are several ways to stay up-to-date with Firefox add-on technology and con
 
 Stay up to date with the latest information about add-on technology on the [Add-ons blog](https://blog.mozilla.org/addons).
 
+### Office hours <span id="office-hours"></span>
```
rebloor, line 56: As this is only for Firefox for Android
```suggestion
### Firefox for Android office hours" <span id="office-hours"></span>
```

rebloor (commented):
Some suggestions

src/content/community.md
dotproto: These suggestions have been integrated.

src/content/community.md
dotproto: This widget has been dropped in response to @Rob--W's feedback. The language used to open the new UI element is "Book a session".

dotproto:
I've replaced the Google Calendar snippet with a `<details>` element and nested `<iframe>`. Before arriving at this solution I explored a few other options including showing the iframe in a `<dialog>` element, the existing `expandable-list` and `expandable-list-item` templates, and enabling Foundation modules that are currently unused such as [Reveal](https://get.foundation/sites/docs/reveal.html). Ultimately using the built in `<details>` element seemed like it would be the fastest solution.

When adding `<details>` styles, I tried to emulate `expandable-list-item`'s appearance. I did not change the way `summary::marker` is displayed because I saw a mention of this having a negative impact on screen readers in Firefox.

The following rule doesn't feel like it should be included here, but I wasn't sure where else to put it or how to better scope it in this project. https://github.com/mozilla/extension-workshop/blob/0b2f4c046096e01e2d075791ed4625918e098016/src/assets/css/_main.scss#L2258-L2265

rebloor (commented):
Text changes LGTM

src/content/community.md
```diff
@@ -52,6 +53,16 @@ There are several ways to stay up-to-date with Firefox add-on technology and con
...
 Stay up to date with the latest information about add-on technology on the [Add-ons blog](https://blog.mozilla.org/addons).
 
+### Firefox for Android office hours <span id="office-hours"></span>
+
+The Firefox Add-ons team holds one-on-one office hours sessions to help you make your extensions available on Android. This is in addition to the [Firefox for Android Add-ons](https://discourse.mozilla.org/c/add-ons/android/9393) forum, where you can seek advice and assistance anytime.
+
+<details>
+  <summary>Book a session</summary>
+
+  <iframe id="booking-frame" src="https://calendar.google.com/calendar/appointments/schedules/AcZssZ3yNnoZNBdJCTiJjqziXs9GoCYpfW9XZFkB0QWKNy9dfoqqPlCw7UsYh48Ff0i3lZcA5YzygwrU?gv=true" sandbox="allow-same-origin allow-scripts allow-forms allow-popups" style="height: 600px;"></iframe>
```
Rob--W, line 63: Just wondering - do we need to add `allow-top-navigation` or even `allow-top-navigation-by-user-activation` as well? Only if a navigation triggered from the calendar is an expected part of the functionality.

And maybe also allow-modals if you want to enable things like alert/confirm.

Rob--W (changes requested):
@dotproto You still need to add the calendar to `frame-src` CSP directive, part of

> In either case, you have to allowlist the frame (and/or script) destination in the CSP. The CSP is configured at
> 
> https://github.com/mozilla/extension-workshop/blob/2e5468e85375dd9e82ebac61a78271cee3f9fc8e/.utils/deploy.sh#L37-L47

FYI: This PR conflicts with #1700; one of the two needs to be rebased depending on whichever lands first.

src/content/community.md
dotproto: The default sandbox value set by the Google Calendar scripts scripts is `"allow-same-origin allow-scripts allow-forms allow-popups allow-popups-to-escape-sandbox allow-storage-access-by-user-activation"`. I tried to trim this list to minimize the capabilities being granted to the sandbox. In retrospect, I _think_ those additional capabilities are reasonable and will therefore go back to the default sandbox settings.

I don't think we need to add navigation-related capabilities since Calendar itself didn't request them.

dotproto:
I did another round of updates and squashed my changes. Please take another look when you're able.

src/content/community.md
```diff
@@ -52,6 +53,32 @@ There are several ways to stay up-to-date with Firefox add-on technology and con
...
+    calendar.schedulingButton.load({
+      url: 'https://calendar.google.com/calendar/appointments/schedules/AcZssZ3yNnoZNBdJCTiJjqziXs9GoCYpfW9XZFkB0QWKNy9dfoqqPlCw7UsYh48Ff0i3lZcA5YzygwrU?gv=true',
+      color: '#039BE5',
+      label: 'Book an appointment',
+      target,
+    });
+  });
+})();
+</script>
+
```
Rob--W, line 81:
```suggestion
```

Rebase went wrong? Since the frame fulfills the role of the scheduler, let's remove the style and script.

Rob--W (changes requested):

src/content/community.md
dotproto: Oh my, yes. Something got off track. I backed up the branch before squashing, I'll get this fixed and re-pushed shortly.

dotproto:
I fixed the bad merge @Rob--W [identified](https://github.com/mozilla/extension-workshop/pull/1698#discussion_r1332790466) and moved the booking iframe styles out of the generic section of the main stylesheet. I'm not sure introducing a new stylesheet called `_pagestyles.scss` is the right approach, but I figured it was cleaner than mixing it in with the general `details` element rules.

Rob--W (approved):
I haven't seen a preview, but the code looks okay.

dotproto:
Merged. This change is now visible on -dev.

PR #1744, opened 2023-11-29
MV3 host_permissions clarification

### LINKED ISSUE #1741: MV3 migration guide should emphasize that host permissions are not granted at install time
The "Host permissions" section of the MV3 migration guide (https://extensionworkshop.com/documentation/develop/manifest-v3-migration-guide/#host-permissions) (and "Migration checklist" at the end of the article) mentions the need to move the origin permissions from `permissions` to `host_permissions`, but fails to mention that (compared to MV2) these permissions are not granted by default (at install time). This causes issues in extensions that expect the permissions to be granted by default. Notably, in Chrome these permissions are still granted by default (but users can revoke the permissions).

The article should be updated to mention the lack of host permissions by default, and link to https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/host_permissions#requested_permissions_and_user_prompts

rebloor:
Add clarification to the Manifest v3 article that `host_permissions` is treated as an optional permission by Firefox. Fixes #1741.


Rob--W (approved):
Thanks!

PR #1758, opened 2023-12-13
Document default build environment that reviewers use for building add-ons in absense of explicit instructions

abhn:
- Explicitly spell out what environment reviewers are using to build add-ons in the absence of details from developers
- Serve as documentation to developers who specify some build instructions but not all (if only the Node version is mentioned, but the developer builds their add-ons on newer Mac devices (arm64) while the reviewer is trying to build it on an x86 device) in case of build mismatches

### DIFF
--- a/src/content/documentation/publish/source-code-submission.md
+++ b/src/content/documentation/publish/source-code-submission.md
@@ -109,7 +109,7 @@ Should any of the above differ from your build environment, it is essential that
 
 {% endcapture %}
 {% include modules/one-column.liquid,
-  id: "default-build-environment"
+  id: "default-reviewer-build-environment"
   content: content
 %}
 
--- a/src/content/documentation/publish/source-code-submission.md
+++ b/src/content/documentation/publish/source-code-submission.md
@@ -105,7 +105,7 @@ Reviewers will try to replicate your build environment if specified. However, if
 - Node 20 LTS and npm 10
 - 35GB of free disk space
 
-Should any of the above differ from your build environment, it is essential that you let the reviewers know about it via the README file in your source submission otherwise the build might not match and your submission may get rejected.
+You must let the reviewers know if any of the above differs from your build environment. Do this in the README file in your source submission. Otherwise, the reviewer's build might not match yours, and your submission may be rejected.
 
 {% endcapture %}
 {% include modules/one-column.liquid,
--- a/src/content/documentation/publish/source-code-submission.md
+++ b/src/content/documentation/publish/source-code-submission.md
@@ -63,14 +63,21 @@ To reproduce the build, the reviewer runs the instructions you provided and then
 - details, including required version and installation instructions, of any tools or utilities that need to be downloaded, for example, [yuicompressor](http://yui.github.io/yuicompressor/).
 - a list of all the commands to generate an identical copy of the extension from the source code, for example, npm install or a grunt target. Ideally, you should include every command in the build script file.
 
+Reviewers will try to replicate your build environment if specified. However, if you do not specify the build environment, here's the default build environment that reviewers use:
+
+- Latest Ubuntu LTS (Desktop edition) virtualized via VirtualBox running on an x64_84 host
+- 10GB of system memory (RAM), 6 cores of vCPU
+- Latest Node LTS and corresponding npm
+- 50GB of disk space
+
+Should any of the above differ from your build environment, it is essential that you let the reviewers know about it via the README file in your source submission otherwise the build might not match and your submission may get rejected.
+
 The tools you use to minify, or concatenate your source code:
 
 - must be open source: we cannot verify a build made with commercial tools.
 - cannot be web-based: all review builds are run locally. Using a web-based tool doesn’t allow the reviewers to be certain that your sources match the minified code. Some web-based tools offer a version that can be run locally, in which case provide a script to run the tool locally.
 
 When using npm, yarn, or other package management tools that support it, be sure to include the lockfile, for example, `package-lock.json`. Otherwise, reviewers may use a different version resulting in differences between the generated code and that in the extension.
-
-Assume the reviewer hasn’t installed any developer tools on their computer, that is, make sure you include all the set-up and build instructions to create your code. However, you don’t need to describe how to install common tools such as npm or node.
 
 Tip: Use a build target relative to the directory containing the source, such as a `dist` subfolder. This makes it easier for the reviewer to locate your extension’s built code.
 
--- a/src/content/documentation/publish/source-code-submission.md
+++ b/src/content/documentation/publish/source-code-submission.md
@@ -63,15 +63,6 @@ To reproduce the build, the reviewer runs the instructions you provided and then
 - details, including required version and installation instructions, of any tools or utilities that need to be downloaded, for example, [yuicompressor](http://yui.github.io/yuicompressor/).
 - a list of all the commands to generate an identical copy of the extension from the source code, for example, npm install or a grunt target. Ideally, you should include every command in the build script file.
 
-Reviewers will try to replicate your build environment if specified. However, if you do not specify the build environment, here's the default build environment that reviewers use:
-
-- Ubuntu 22.04 LTS (Desktop edition)
-- 10GB of system memory (RAM), 6 cores of vCPU
-- Node 20 LTS and npm 10
-- 35GB of free disk space
-
-Should any of the above differ from your build environment, it is essential that you let the reviewers know about it via the README file in your source submission otherwise the build might not match and your submission may get rejected.
-
 The tools you use to minify, or concatenate your source code:
 
 - must be open source: we cannot verify a build made with commercial tools.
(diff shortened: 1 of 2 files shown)


wagnerand:
@dotproto could you give this a look as well, from a developer relations engineering perspective?

src/content/documentation/publish/source-code-submission.md
```diff
@@ -63,15 +63,22 @@ To reproduce the build, the reviewer runs the instructions you provided and then
 - details, including required version and installation instructions, of any tools or utilities that need to be downloaded, for example, [yuicompressor](http://yui.github.io/yuicompressor/).
 - a list of all the commands to generate an identical copy of the extension from the source code, for example, npm install or a grunt target. Ideally, you should include every command in the build script file.
 
+Reviewers will try to replicate your build environment if specified. However, if you do not specify the build environment, here's the default build environment that reviewers use:
```
wagnerand, line 66: Could we add a subheading here? Would be useful for when we include the link to this section in the canned response.

src/content/documentation/publish/source-code-submission.md
```diff
@@ -63,15 +63,22 @@ To reproduce the build, the reviewer runs the instructions you provided and then
...
+
+- Latest Ubuntu LTS (Desktop edition) virtualized via VirtualBox running on an x86_64 host
```
wagnerand, line 68: Let's spell out the version here explicitly as well, to be as clear as possible to developers.

We should also remove the host information. It should not be relevant, and might not be true for everyone.

src/content/documentation/publish/source-code-submission.md
```diff
@@ -63,15 +63,22 @@ To reproduce the build, the reviewer runs the instructions you provided and then
...
+- 10GB of system memory (RAM), 6 cores of vCPU
+- Latest Node LTS and corresponding npm
```
wagnerand, line 70: Same here, let's add specific (major) version numbers.

src/content/documentation/publish/source-code-submission.md
```diff
@@ -63,15 +63,22 @@ To reproduce the build, the reviewer runs the instructions you provided and then
...
+- 50GB of disk space
```
wagnerand, line 71: Is that free disk space after the OS has been installed?

src/content/documentation/publish/source-code-submission.md
```diff
@@ -63,15 +63,22 @@ To reproduce the build, the reviewer runs the instructions you provided and then
...
+
+Should any of the above differ from your build environment, it is essential that you let the reviewers know about it via the README file in your source submission otherwise the build might not match and your submission may get rejected.
```
wagnerand, line 73: Just to double-check: Is the README the (one) location we want that to be provided in?

wagnerand (changes requested):

dotproto:
We may also want to update https://wiki.mozilla.org/Add-ons/Reviewers/Guide/Reviewing to either cross link or include their info included in this change.

src/content/documentation/publish/source-code-submission.md
abhn: I was contemplating whether to add "Notes for reviewers" here, but didn't see it mentioned for anything source code submission related from before so didn't add. Ideally the reviewer should be made aware of the platform requirements before they have already downloaded the source code inside their VM, which makes "Notes for reviewers" a slightly better place for it. Thoughts on if we change our recommendation from README to "Notes for reviewers" or something better? 

FWIW, we do mention 3PL sources be added to "Notes for reviewers"

wagnerand:
> We may also want to update [wiki.mozilla.org/Add-ons/Reviewers/Guide/Reviewing](https://wiki.mozilla.org/Add-ons/Reviewers/Guide/Reviewing) to either cross link or include the info included in this change.

The reviewer documentation for default build environments currently lives [here](https://mozilla-hub.atlassian.net/wiki/spaces/FDPDT/pages/435224648/Default+build+environment+for+AMO+reviewers). Long term, I think there is value in having one source of truth if possible. While I don't expect that, with private-first documents, we need to check whether there is any information that should not be public.
It's not a big priority, but something we can do at some point, along with updating that public documentation in general.

src/content/documentation/publish/source-code-submission.md
abhn: Good one, thanks!

src/content/documentation/publish/source-code-submission.md
wagnerand: I don't have a preference, we should pick what works best for reviewers (from the available options of course, we don't want to add an additional burden to developers by coming up with something entirely new). For me, it's more important that there is **exactly one (!)** (also not zero) location that both the developer and reviewers are aligned on, rather than forcing a particular location (until we have a dedicated way to provide that).

I personally like when all information is within the source zip, because it's immutable and contained with the code itself, rather than in a different place, but that is not necessarily the ideal state, as developers might need to update their instructions sometimes. Having to resubmit the entire source package in that case might be beneficial overall to ensure code and instructions are somewhat in sync but it is also more effort than to just update a text field on AMO.
However, if instructions are in multiple places, they tend to be out of sync and that's when ambiguity, inaccuracy and conflicting information lead to reproducing failures, rejections and back and forth with the developer.

wagnerand: @dotproto for context, there is an [idea](https://mozilla-hub.atlassian.net/browse/AOP-154) to standardize this, which to me sounds valuable to do.

abhn: I agree that while README isn't ideal, it does come with its own merits and I'd favor not changing it right now (it does come with the additional work for reviewers to switch environment after opening up the source package, but this PR doesn't affect/change that effort). Let's stick to README and we can work towards [standardizing it](https://mozilla-hub.atlassian.net/browse/AOP-154).

src/content/documentation/publish/source-code-submission.md
dotproto: This was addressed in commit 6940b3b.

src/content/documentation/publish/source-code-submission.md
```diff
@@ -99,6 +97,24 @@ You can add source code to an existing add-on version. To do this, open [My Add-
 
 {% capture content %}
```
wagnerand, line 98: [Github getting the diff wrong, this would be part of your patch]

I don't really know how things work, but are you sure this is working? There is another section called `content` already defined in another part of this document. Can a page have the different sections with the same name?

src/content/documentation/publish/source-code-submission.md
dotproto: Thanks for the additional context, @wagnerand.

src/content/documentation/publish/source-code-submission.md
dotproto: Here `content` is essentially a variable for the template system, so it can be reused across the different sub-sections of the document. For linking purposes the important part is that we have different IDs assigned to each of these sections (for example, [this line](https://github.com/mozilla/extension-workshop/pull/1758/files/4e18ce5ea3c567f101c651a62f29067a711f5e1f#diff-362cb30ebfed8b5d820e5364d8b8c3120d7278bf64469dc1eb9cc49311c5c157R112)).

wagnerand: Thank you.

abhn: (was put off by the same thing but looked up at other places and then understood hehe 🙃)

src/content/documentation/publish/source-code-submission.md
```diff
@@ -99,6 +97,24 @@ You can add source code to an existing add-on version. To do this, open [My Add-
...
 
+### Default reviewer build environment
+Reviewers will try to replicate your build environment if specified. However, if you do not specify the build environment, here's the default build environment that reviewers use:
+
+- Ubuntu 22.04 LTS (Desktop edition)
+- 10GB of system memory (RAM), 6 cores of vCPU
+- Node 20 LTS and npm 10
+- 35GB of free disk space
```
dotproto, line 106: In [previous](https://github.com/mozilla/extension-workshop/pull/1758#discussion_r1425550871) [comments](https://github.com/mozilla/extension-workshop/pull/1758#discussion_r1425552660) @wagnerand suggested we include specific version numbers to remove any potential ambiguity. While I agree with the motivation behind this suggestion, I'm hesitant about the maintenance overhead that explicit version numbers creates. 

As an alternative, what if we said something like "The latest Node.js LTS release" and link this to either the [Node.js download page](https://nodejs.org/en) or the documentation on [previous releases](https://nodejs.org/en/about/previous-releases#looking-for-latest-release-of-a-version-branch). Similarly for Ubuntu we could link to the Ubuntu website's [downloads page](https://ubuntu.com/download), [releases page](https://www.releases.ubuntu.com/), or the wiki's [releases page](https://wiki.ubuntu.com/Releases).

wagnerand: There is definitely a little bit of overhead involved updating this page. However, we don't plan to re-create the VM very often, mostly for major updates for ubunut and node, maybe python which happen only every few years.
The bigger risk I see is us forgetting to update this page when we re-recreate the VM, which is why I [asked](https://mozilla-hub.atlassian.net/wiki/spaces/FDPDT/pages/435224648/Default+build+environment+for+AMO+reviewers?focusedCommentId=493813858) that we add an "update documentation page" step.

Adding links either way sounds great.

rebloor (commented):
Suggested edits for your consideration

wagnerand (approved):

wagnerand:
Merged on @abhn's request. All reviewer comments have been addressed.

PR #1783, opened 2024-01-19
`web-ext run --arg` - Mention `--remote-debugging-port`

fregante:
I'm trying to debug a crashing background page extension, I had a hard time finding `--arg` as a way to remotely debug the background page. Given it's a _relatively_ common flag, I think it would be good to mention it for discoverability.


willdurand (approved):

willdurand:
thanks

PR #1785, opened 2024-01-29
Allow enterprise extensions to use storage.managed to grant consent.

mkaply:


dotproto (approved):
I _think_ this is sufficient for an initial update but I would also like to have @wagnerand approve before merging. 

We should follow up this PR with more detailed guidance for developers. I've opened #1786 to track that work.

src/content/documentation/develop/best-practices-for-collecting-user-data-consents.md
```diff
@@ -20,6 +20,8 @@ If your extension collects user data, it needs to comply with the requirement of
 
 A common developer question about these policies is how to translate them into browser extension features that are compliant to [Mozilla’s Add-on policies](/documentation/publish/add-on-policies/). This how-to is the result of those requests and offers advice on implementing prompts to meet the data collection and add-on policies. This article suggests how you can implement suitable prompts but it doesn’t replace or supersede the policies; you still need to confirm that your extension complies with the policies.
 
+Note: For extensions managed by enterprise policy, consent may be granted outside of the extension. Policy-managed extensions may use storage.managed to detect such a grant and skip the consent confirmation step.
```
wagnerand, line 23: How do we define "extensions managed by enterprise policy"?

wagnerand, line 23: > ...consent may be granted outside of the extension. Policy-managed extensions may use storage.managed to detect such a grant and skip the consent confirmation step.

Can we rephrase this to clarify what it means and what the requirements are? This is using `may` a lot, some things should probably use `must`. For example, is it clear that using `storage.management` is the **only** (approved) way to bypass the disclosure and consent experience?

wagnerand, line 23: > ... and skip the consent confirmation step.

Let's make it clear and add a note that the disclosure and consent experience still needs to be implemented in the add-on and shown when the respective `storage.managed` key is not set.

wagnerand (changes requested):

mkaply:
New patch pushed with updated text approved by Andreas.

wagnerand (approved):

PR #1797, opened 2024-02-15
Fix existing redirects

### LINKED ISSUE #1796: Redirects are no longer working

Go to any of the redirects we set up in https://github.com/mozilla/extension-workshop/pull/1563, for example: https://extensionworkshop.com/documentation/develop/developing-extensions-for-firefox-for-android-fennec/

### What happened?

404

Not a 404 but a redirect.

willdurand:
Fixes #1796


diox (approved):
Verified that it matches the redirects we're setting up, but it's difficult to test without deploying to some S3 bucket for real.

PR #1841, opened 2024-04-23
Update GitHub Actions workflow to trigger prod deploys and build version.json correctly

pattersonbl2:
prod changes:
adding new tag regex to prod 
removing immutable flag from build step
formatting fixes 
change fetch-depth to get all history 

stage changes: 
change fetch-depth to get all history

### DIFF
--- a/.github/workflows/gcp_deploy.yaml
+++ b/.github/workflows/gcp_deploy.yaml
@@ -20,7 +20,7 @@ jobs:
           node-version: "18"
 
       - name: Install dependencies
-        run: yarn install
+        run: yarn install --immutable
 
       - name: Test current build
         run: yarn run test-ci
--- a/.github/workflows/gcp_deploy.yaml
+++ b/.github/workflows/gcp_deploy.yaml
@@ -11,6 +11,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - name: Setup Node.js
         uses: actions/setup-node@v3
@@ -18,7 +20,7 @@ jobs:
           node-version: "18"
 
       - name: Install dependencies
-        run: yarn install --immutable
+        run: yarn install
 
       - name: Test current build
         run: yarn run test-ci
@@ -33,8 +35,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 1
-          fetch-tags: true
+          fetch-depth: 0
 
       - name: Setup Node.js
         uses: actions/setup-node@v3
--- a/.github/workflows/gcp_deploy_prod.yaml
+++ b/.github/workflows/gcp_deploy_prod.yaml
@@ -3,7 +3,7 @@ name: Deploy to GCS Prod
 on:
   push:
     tags:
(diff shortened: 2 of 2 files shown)


.github/workflows/gcp_deploy_prod.yaml
```diff
@@ -50,7 +52,7 @@ jobs:
 
       - name: Build the website
         run: |
-          yarn install --immutable
+          yarn install
```
willdurand, line 55: Let's keep `--immutable` here and above

.github/workflows/gcp_deploy_prod.yaml
```diff
@@ -3,7 +3,7 @@ name: Deploy to GCS Prod
 on:
   push:
     tags:
-      - "20[0-9][0-9].[0-9][0-9].[0-9][0-9](-[0-9]+)?"
+      - "202[0-9].[0-9][0-9].[0-9][0-9]"
```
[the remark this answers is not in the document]
pattersonbl2: The previous way didn't trigger the action like i wanted but i was able to trigger it  by adding a * to the tag regex.  I have just pushed up this change.

.github/workflows/gcp_deploy_prod.yaml
pattersonbl2: added both back.

.github/workflows/gcp_deploy_prod.yaml
```diff
@@ -3,7 +3,7 @@ name: Deploy to GCS Prod
 on:
   push:
     tags:
-      - "20[0-9][0-9].[0-9][0-9].[0-9][0-9](-[0-9]+)?"
+      - "202[0-9].[0-9][0-9].[0-9][0-9]*"
```
willdurand, line 6: Maybe simplify to allow any tag that starts with `20`. We don't need to be very precise here, I think.

```suggestion
      - "20*"
```

.github/workflows/gcp_deploy_prod.yaml
```diff
@@ -50,7 +52,7 @@ jobs:
 
       - name: Build the website
         run: |
-          yarn install --immutable
+          yarn install  --immutable
```
willdurand, line 55:
```suggestion
          yarn install --immutable
```

.github/workflows/gcp_deploy.yaml
```diff
@@ -11,14 +11,16 @@ jobs:
...
+          fetch-depth: 0
 
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
           node-version: "18"
 
       - name: Install dependencies
-        run: yarn install --immutable
+        run: yarn install
```
willdurand, line 23:
```suggestion
        run: yarn install --immutable
```

.github/workflows/gcp_deploy_prod.yaml
pattersonbl2: Long as the same tag is being pushed and no one pushes anything not following the pattern

pattersonbl2:
@willdurand  i have commit all the suggestion.  if it's ok i would like to merge this in today so i can start planning the move to prod.

willdurand (approved):

PR #1844, opened 2024-04-27
Correcting a typo in "submitting-an-add-on.md"

achangofpace:
Extensin


rebloor (approved):
Thanks for spotting that one @achangofpace

PR #1865, opened 2024-05-23
Fix title for the web-ext v7 reference page

willdurand:
When we type "web-ext" in the search bar, we get two results and both have the same title:

Screenshot 2024-05-23 at 14 34 55

This PR addresses that.


dotproto (approved):
Thanks, @willdurand!

Issue #1866, opened 2024-05-23
`--firefox-preview` has been removed in web-ext v8


willdurand:
See: https://github.com/mozilla/web-ext/issues/3104

PR #1867, opened 2024-05-24
--firefox-preview is removed in web-ext v8

willdurand:
Fixes https://github.com/mozilla/extension-workshop/issues/1866


rebloor (approved):
@willdurand I've removed the relevant documentation too. You might want to double check your local preview, but otherwise looks good.

PR #1886, opened 2024-06-12
Fallback for _execute_browser_action

rebloor:
### Description

Addresses the documentation needed for [Bug 1797811](https://bugzilla.mozilla.org/show_bug.cgi?id=1797811) Migrate user-defined shortcuts from _execute_browser_action to _execute_action, by adding a note to the Manifest v3 migration guide.

### Related issues and pull requests

Related update to MDN https://github.com/mdn/content/pull/34091

PR #1898, opened 2024-06-26
Action retention

rebloor:
Remove comment suggesting that action API, key, and special shortcut will be removed. See. https://github.com/mozilla/extension-workshop/pull/1886/files#r1652208877


Rob--W:
Thanks for this PR. I'll confirm with my team whether non-deprecation of page_action is intentional and then sign off if it is.

The relevant bug tracking the action API consolidation is https://bugzilla.mozilla.org/show_bug.cgi?id=1687756

Rob--W (approved):

PR #1994, opened 2025-01-20
Include Markdown Syntax Information

chrstinalin:
Adds Markdown syntax information for https://github.com/mozilla/addons-server/pull/22956, to then be linked to.

image

### DIFF
--- a/src/content/documentation/publish/create-an-appealing-listing.md
+++ b/src/content/documentation/publish/create-an-appealing-listing.md
@@ -293,6 +293,86 @@ And, when users suggest improvements or highlight issues in the reviews, use the
 
 <!-- Single Column Body Module -->
 
+<section id="make-use-of-markdown" class="module">
+<article class="module-content grid-x grid-padding-x">
+<div class="cell small-12">
+
+## Make use of Markdown
+
+Some fields in your add-on's listing support a limited set of Markdown.
+<table>
+   <thead>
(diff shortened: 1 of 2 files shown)


src/content/documentation/publish/create-an-appealing-listing.md
```diff
@@ -293,6 +293,86 @@ And, when users suggest improvements or highlight issues in the reviews, use the
 
 <!-- Single Column Body Module -->
 
+<section id="make-use-of-markdown" class="module">
+<article class="module-content grid-x grid-padding-x">
+<div class="cell small-12">
+
+## Make use of Markdown
+
+Some fields in your add-on's listing support a limited set of Markdown.
```
rebloor, line 302: Can we be any more specific amount which fields these are? Otherwise this leave the reader to experiment and discover for themselves.

KevinMind (approved):
I would love to see an additional snippet which serves as a "copy paste" snippet exemplifying all of the supported syntax. This would enable a dev to simply copy past the snippet onto amo and then immediately start formatting and shaping the content.

I've found this super useful in the past and would love to see it here. Otherwise looks good.

I don't see any testing instructions so assuming there is nothing to verify?

chrstinalin:
> I don't see any testing instructions so assuming there is nothing to verify?

Nope, just the text content/formatting.

PR #1997, opened 2025-01-23
Switch from node-sass (which is EOL) to sass-embedded

diox:
Ref https://github.com/mozilla/addons/issues/15273


eviljeff (approved):
no incompatibilities so far!

diox:
extension-workshop was the easy one, there was an incompatibility in the way we defined breakpoints on addons-blog and addons-frontend. Some deprecation warnings though, but that requires a lot of changes so I've ignored those for now.

PR #2036, opened 2025-03-19
Instructions for sideloading on Firefox for Android

### LINKED ISSUE #2028: "Self Distribution" page should state that Android cannot directly install self-distributed extensions
User [robinmetral](https://discourse.mozilla.org/u/robinmetral) on Discourse [shared](https://discourse.mozilla.org/t/why-can-i-not-load-a-simple-custom-extension-addon-xpi-file-in-android-nightly/129335/3) the following suggestion:

> > At the moment Firefox for Android only supports installing add-ons directly from [addons.mozilla.org](http://addons.mozilla.org/); attempting to install an add-on from another site will end up downloading the XPI file.
>
> Is this documented anywhere? It would be very useful as an aside on Extension Workshop’s [Self distribution page](https://extensionworkshop.com/documentation/publish/self-distribution/), for example. The current version of the page makes it seem like the only “downsides” to self distribution are lack of visibility and the need to publish an update manifest manually. No out-of-the-box availability on Android is kind of a big deal.

Request: Update the Self Distribution to note that on Android following a link to an `.xpi` file will download the file rather than install it as on Desktop.

rebloor:
- Removes the term sideloading (and replaces with "manual installation")
- Adds instructions to the (now) installing self-distributed add-ons page on how to activate and use **Install Extension from File** on Firefox for Android. 
- Completed various minor edits to the affected pages.

Fixes #2028


src/content/documentation/publish/distribute-sideloading.md
dotproto, line 1: The naming of this article is unfortunate as this term has a specific meaning. As described in [Support for extension sideloading has ended](https://blog.mozilla.org/addons/2020/03/10/support-for-extension-sideloading-has-ended/), "sideloading" generally refers to an installation method where installing an application on your system would also result in adding and extension to Firefox. 

The "Installation using the standard extension folders" section does call out that this feature is no longer supported, but the intro still seems to suggest it is supported.

src/content/documentation/publish/distribute-sideloading.md
```diff
@@ -15,13 +15,13 @@ date: 2019-11-06 04:05:42
...
-You may want to send a user your add-on’s XPI file by some means other than a web download, such as an email distribution of a beta version for user testing. In this case, there are two practical options for installing the add-on:
+You may want to send a user your add-on’s XPI file by some means other than a web download, such as an email distribution of a beta version for user testing. In this case, there are two options for installing the add-on:
 
-- Using **Install Add-on From File** in the Add-ons Manager.
+- Installing an extension from file in desktop Firefox or Firefox for Android.
 - Adding the file to one of the standard extension folders.
 
 ::: note
-No automatic updates will be performed for add-ons installed using these methods. You will have to deliver a new XPI file to your user for each update. However, automatic compatibility checks are still performed.
+No automatic updates are performed for add-ons installed using these methods. You have to deliver a new XPI file to your user for each update. However, automatic compatibility checks are performed.
```
dotproto, line 24: I don't think this is accurate. I tried walking through extensions source files on searchfox but I didn't see anything that suggested updates are handled different based on installation method.

@willdurand, can you confirm?

src/content/documentation/publish/distribute-sideloading.md
```diff
@@ -67,25 +67,25 @@ An email address style ID is recommended.
 
 {% capture content %}
 
-## Install add-on from file
+## Install add-on from file on a computer
 
-To use Install Add-on From File in Add-on Manager, send the user the signed add-on with the following instructions:
+To sideload an add-on in desktop Firefox, send the user the signed add-on with these instructions:
```
dotproto, line 72: We should avoid the term "sideload" here as that has a more specific meaning.

```suggestion
To install an add-on from a file in desktop Firefox, send the user the signed add-on with these instructions:
```

dotproto (changes requested):

src/content/documentation/publish/distribute-sideloading.md
rebloor: @dotproto Isn't sideloading a general term meaning  "installing applications or software on a device from sources other than the official app store or distribution channels, bypassing the usual vetting process". The feature removed was the ability to sideload "without the user taking an action"?

rebloor: @dotproto I have revisited and removed the information about installing with an app. Given this was removed five years ago, there seems a little point in including the documentation.

src/content/documentation/publish/distribute-sideloading.md
rebloor: @dotproto after reviewing the documentation, I believe this statement was actually intended to refer to extensions offered as a beta update release to a listed extension. I've updated the note.

rebloor:
@dotproto @willdurand I have now adjusted the content to remove the term sideloading. I've also renamed the page's md file and created https://github.com/mozilla/extension-workshop/issues/2045 for the introduction of a redirect.

src/content/documentation/publish/self-distribution.md
```diff
@@ -89,15 +98,15 @@ When distributing an add–on from your website, you have two options for making
...
   </div>
   ```
-  If you take this approach, make sure that the browser detects your JavaScript code as handling user input, for example, it is called from inside a DOM click event.
+  If you take this approach, make sure that the browser detects your JavaScript code as handling user input. For example, it is called from inside a DOM click event.
 
 If the download link or triggering JavaScript is not on the top-level frame of the website, none of the frames between the top-level page and the frame containing the download or JavaScript can be third-party or cross-origin frames.
 
 ::: note
-You may find references to using the `InstallTrigger` API to trigger a download from JavaScript. This API was never officially supported and will be removed from Firefox 102. Use is therefore not recommended.
+You may find references to using the `InstallTrigger` API to trigger a download from JavaScript. This API was never officially supported and will be removed from Firefox 102. Use is, therefore, not recommended.
```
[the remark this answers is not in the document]
rebloor: @dotproto it's not been removed, and the relevant bug is still open.
```suggestion
You may find references to using the `InstallTrigger` API to trigger a download from JavaScript. This API is not supported and it will be removed (see [Bug 1776426](https://bugzilla.mozilla.org/show_bug.cgi?id=1776426)).  Use is, therefore, not recommended.
```

PR #2049, opened 2025-04-10
Add anchors to policy sub-sections

dotproto:
This PR addresses a long standing frustration of mine: it's not possible to link developers directly to a sub-section of the add-ons policies. To address this, I've manually hacked in `<span>` tags with ID attributes to the H3 and H4 elements on the policies page.

Ideally we would have a more robust way of generating header anchors but for now this is better than nothing.

### DIFF
--- a/src/content/documentation/publish/add-on-policies.md
+++ b/src/content/documentation/publish/add-on-policies.md
@@ -13,9 +13,10 @@ contributors:
     jvillalobos,
     wbamberg,
     kmaglione,
+    dotproto
   ]
-last_updated_by: wagnerand
-date: 2024-02-15
+last_updated_by: dotproto
+date: 2025-04-10
 ---
 
 <!-- Page Hero Banner -->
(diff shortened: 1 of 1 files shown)


src/content/documentation/publish/add-on-policies.md
```diff
@@ -45,7 +46,7 @@ When an add-on is given human review or otherwise assessed by Mozilla, these pol
 
 Users should be able to easily discern the functionality of your add-on based on the listing, and should not be presented with unexpected user experiences after installing it. The listing should include an easy-to-read description of what the add-on does, and what information it collects. Please consult our best practices guide for [creating an appealing listing](/documentation/develop/create-an-appealing-listing/).
 
-### Unexpected features
+### <span id="unexpected-features"></span>Unexpected features
```
rebloor, line 49: @dotproto I assume you've tested and these work as expected? I ask, as it appears that the "correct" way of doing this would be like this, as per `develop.md`:
```
<section id="user-experience" class="page-section-container">

<!-- Single Column Body Module -->

{% capture content %}

## User Experience

{% endcapture %}
{% include modules/one-column.liquid,
	content: content
%}

<!-- END: Single Column Body Module -->
```

willdurand:
@dotproto what is the status of this PR?

wagnerand:
Perhaps this should be closed in favor of https://github.com/mozilla/extension-workshop/pull/2077 which would give us that for free.

willdurand:
I agree.

PR #2052, opened 2025-04-21
Add default values for timeouts to web-ext docs

jellyfith:
I was debugging a timeout issue on one of my extensions and noticed that mozilla provides a few timeout configuration options but doesn't list the default values, which could be good for posterity.


jellyfith:
@dotproto 👋 What is the best way to get eyes on a ticket like this, without a ticket attached?

willdurand (approved):

willdurand:
Thanks!

PR #2022, opened 2025-03-04
Clarify extension ID creation

andrewbaxter:
AFAICT nowhere in the docs did it say that the developer can come up with the extension ID on their own, and in several places it's suggested that they don't come up with it on their own (ex: testing locally, uploading manifest V2 extensions with no ID).

For manifest V3, the developer must come up with an ID on their own to submit the extension. This confused me for several hours until I joined the addons web chat and got an answer from someone official.  

In this MR I clarified it (I think in the best location) and provided details on the requirements and how to to choose one.

Related: https://github.com/mdn/content/pull/38444 - I also made changes here because this information is spread in multiple places so people might not find one or the other, and without this clarification reading the wrong document could lead someone to think they have the (wrong) answer and stop searching at that point.

PR #2125, opened 2025-07-22
Use `@name` formatting for add-on IDs

dotproto:
Historically add-ons documentation has used email address-style ID names. This has created confusion among developers and lead to a variety of incorrect assumptions about the requirements, behaviors, limitations, and validation steps associated with these IDs.

In order to sidestep the confusion created by this format, this commit replaces replaces the email convention (e.g. addon-name@example.com) with a simpler @-prefixed ID (e.g. @example). This should better communicate that the ID needs to be a unique value in a global namespace while still enabling developers to choose human-readable names for their add-ons.

### DIFF
--- a/src/content/content-guidelines/example.md
+++ b/src/content/content-guidelines/example.md
@@ -118,7 +118,7 @@ Take the popular extension, Remembear, as an example. You can imagine the creato
 
 #### 2. Identify key words
 
-Pull out the key words (nouns, adjectives, verbs) and phrases from your description and create a word list. Add other words that come to mind based on those words, including synonyms.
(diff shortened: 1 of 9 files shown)


dotproto:
#2022 has been merged and this PR is has been rebased on top of the updated on top of the [master](https://github.com/mozilla/extension-workshop/tree/master) branch.

I also updated the template to replace inline HTML with the appropriate set of liquid templates. This addresses a minor issue where this page's table of contents appeared halfway the page rather than below the header.

rebloor (approved):
@dotproto change you look good to me. Are you going to make the parallel changes to MDN or would you like me to do that?

dotproto:
@rebloor, if you could, that would help quite a bit. Hm, I guess we should also apply this to the examples repo. I'll start looking at that now.

PR #2129, opened 2025-07-22
Clarify how other browsers handle the BSS key

### LINKED ISSUE #2126: Remove mention of Chrome warning on BSS in the add-on ID docs
@dotproto I've taken one of the Chrome extension examples and added `browser_specific_settings` to it, then loaded it as an unpacked file. Chrome reported no errors or warnings. Can you confirm the statement about a warning in Chrome for `browser_specific_settings` is incorrect/updated?

_Originally posted by @rebloor in https://github.com/mozilla/extension-workshop/pull/2022#discussion_r2029931355_

dotproto:
Chrome stopped logging a warning when the `browser_specific_settings` key is present in an extensions' manifest. This PR updates the note in our add-on ID docs to clarify Chrome and Safari behavior when parsing this object. 

Fixes #2126

### DIFF
--- a/src/content/content-guidelines/example.md
+++ b/src/content/content-guidelines/example.md
@@ -118,7 +118,7 @@ Take the popular extension, Remembear, as an example. You can imagine the creato
 
 #### 2. Identify key words
 
-Pull out the key words (nouns, adjectives, verbs) and phrases from your description  and create a word list. Add other words that come to mind based on those words, including synonyms.
+Pull out the key words (nouns, adjectives, verbs) and phrases from your description and create a word list. Add other words that come to mind based on those words, including synonyms.
 
 Remembear Example:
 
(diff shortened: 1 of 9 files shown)


rebloor (commented):
@dotproto I'm wondering whether we need to re-factor this content to make clear that this applies to both manifest V2 and V3 (where you have to specify an ID and therefore include `browser_specific_settings`). In addition, we should probably mention this in the cross browser discussion e.g. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Build_a_cross_browser_extension#manifest_keys

dotproto:
@rebloor are we good to merge this change?

rebloor:
@dotproto if you're happy that we don't need to do anything re-my comment, then yes.
> I'm wondering whether we need to re-factor this content to make clear that this applies to both manifest V2 and V3 (where you have to specify an ID and therefore include `browser_specific_settings`). In addition, we should probably mention this in the cross-browser discussion e.g., https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Build_a_cross_browser_extension#manifest_keys

PR #2133, opened 2025-08-01
Turn the policies update preview into effect

wagnerand:
This moves the previews into a permanent state, with the policies update going into effect on Aug 4. To align effectiveness with that date, I intentionally set the `date` to Aug 4.


src/content/documentation/publish/add-on-policies-preview-2025-08.md
```diff
@@ -1,303 +0,0 @@
----
```
willdurand, line 1: I think we need to keep this URL since it has been shared previously. What if we start by keeping the content, and we just say something like:

> The policies are available at: <link>

willdurand (approved):

PR #2167, opened 2025-09-16
Issue 24191 Advice on bugging "can't access dead object" errors

### LINKED ISSUE #1566: Add details about "can't access dead object" in debugging
See  incorrect links in "can't access dead object" page [#24191](https://github.com/mdn/content/issues/24191)

rebloor:
Add a section to the deep bugging page that provides guidance on resolving "can't access dead object" errors.

Fixes #1566, which in turn addresses issues originally raised in https://github.com/mdn/content/issues/24191.

### DIFF
--- a/src/content/documentation/develop/debugging.md
+++ b/src/content/documentation/develop/debugging.md
@@ -432,6 +432,33 @@ For more details, see [Testing persistent and restart features](/documentation/d
 
 {% capture content %}
 
+## Resolving "can't access dead object" errors
+
+Extensions are prevented from keeping references to DOM objects after their parent document is destroyed. This prevents memory leaks in extensions.
+
+If your extension references a nonexistent DOM object, you get an error message similar to this in the Console:
+
+```
+Error: TypeError: can't access dead object
+Source File: chrome://addon/content/file.js
+Line: 42
+```
+
+To correct this …
+
+{% endcapture %}
+{% include modules/one-column.liquid,
+  id: "cant-access-dead-object"
+  content: content
+%}
+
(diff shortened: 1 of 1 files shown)


rebloor:
@dotproto can you offer any suggestions on how a developer would go about resolving this error? (e.g., something more specific than identify and prevent access to the DOM objects.)

dotproto:
A couple of options leap to mind:

1. Replace the original element/object with a clone
    ```js
    // Old code
    console.log(myElement);
    console.log(myObject);
    
    // New code
    console.log(myElement.cloneNode());
    consoe.log(structuredClone(myObject));
    ```
2. Instead of logging a specific object that could mutate or be deleted after the log statement, stringify the object (or a subset of the object)
    ```js
    // Old code
    console.log(object);
    
    // New code
    console.log(JSON.stringify(object));
    ```

dotproto:
There's way too much subtly in my last comment. I think the simplest, most actionable guidance we can share is to recommend that developers only log the specific properties they need where possible. If they need to log objects during development, they can use `element.cloneNode()` for elements and `structuredClone(object)` for JavaScript objects, but this should be avoided (especially in production) due to the additional memory overhead this can incur.

rebloor:
@dotproto is this just about logs? The [original blog post](https://blog.mozilla.org/addons/2012/09/12/what-does-cant-access-dead-object-mean/) suggested that this issue could prevent the extension from working correctly. Would failing to log information from a dead object have the same effect?

dotproto (approved):
Sounds like a plan. Thanks, @rebloor!

PR #2175, opened 2025-09-23
Updated Android debugging advice

rebloor:
This change updates the details about debugging on an Android device:

- updates the recommended base version of web-ext
-  removes the advice about using Nightly only
- list the paired computer and Android editions
- notes that versions don't have to be the same, but should be aligned if issues are encountered
- add the APK names for each edition
- moves cross-reference links to Firefox source docs to the setup section and points back to the set up section from Debug your extension rather than using external links
- other minor grammar changes

### DIFF
--- a/src/content/documentation/develop/developing-extensions-for-firefox-for-android.md
+++ b/src/content/documentation/develop/developing-extensions-for-firefox-for-android.md
@@ -47,7 +47,7 @@ Complete some one-off setup tasks on your computer and Android device.
 
 On your development computer:
 
-- Install or update [web-ext](https://github.com/mozilla/web-ext) to version 7.12.0 or later.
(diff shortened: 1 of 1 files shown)


src/content/documentation/develop/developing-extensions-for-firefox-for-android.md
```diff
@@ -47,16 +47,24 @@ Complete some one-off setup tasks on your computer and Android device.
 
 On your development computer:
 
-- Install or update [web-ext](https://github.com/mozilla/web-ext) to version 4.1.0 or later.
+- Install or update [web-ext](https://github.com/mozilla/web-ext) to version 7.12.0 or later.
```
Rob--W: For future reference, the 7.12 recommendation here follows from the fix needed to support current Android versions, implemented in https://github.com/mozilla/web-ext/issues/2312

PR #2176, opened 2025-09-24
Recommend ES modules in the web-ext config section

willdurand:


src/content/documentation/develop/getting-started-with-web-ext.md
```diff
@@ -526,7 +527,7 @@ You can specify `--config=my-config.cjs` or `--config=my-config.mjs` to set defa
 web-ext --config=my-config.mjs build
 ```
 
-The file should be a CommonJS module [as understood by NodeJS](https://nodejs.org/docs/latest/api/modules.html#modules_modules) and must export each configuration value. Here is how you would set the default value of [--verbose](/documentation/develop/web-ext-command-reference/#verbose) to `true`:
+The file should be an ES module [as understood by Node.js](https://nodejs.org/docs/latest/api/esm.html) and must export each configuration value. Here is how you would set the default value of [--verbose](/documentation/develop/web-ext-command-reference/#verbose) to `true`:
```
willdurand, line 530: This is the only sentence that I actually changed. The rest is extra spaces removed automatically by my `vim`.

rpl (approved):
lgtm, Thanks @willdurand

PR #2203, opened 2025-10-29
Fix "Manifiest" typo

dotproto:
The add-on ID documentation had two occurrences of the typo "Manifiest" in place of "Manifest". This PR corrects that.


rebloor:
Thanks @dotproto

PR #2269, opened 2026-03-02
web-ext-command-reference: clarify that build outputs to artifacts dir

Zocker1999NET:
So that when only reading web-ext build you are guided to the existing global option

### DIFF
--- a/src/content/documentation/develop/web-ext-command-reference.md
+++ b/src/content/documentation/develop/web-ext-command-reference.md
@@ -147,7 +147,7 @@ Environment variable: `$WEB_EXT_OVERWRITE_DEST=true`
 
 #### `--filename`, `-n`
 
-Name of the created extension package file. In this option, the values defined in `manifest.json` can be used by enclosing them with `{ }`. The default value is `{name}-{version}.zip`.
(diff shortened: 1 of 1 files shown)


src/content/documentation/develop/web-ext-command-reference.md
```diff
@@ -147,7 +147,7 @@ Environment variable: `$WEB_EXT_OVERWRITE_DEST=true`
 
 #### `--filename`, `-n`
 
-Name of the created extension package file. In this option, the values defined in `manifest.json` can be used by enclosing them with `{ }`. The default value is `{name}-{version}.zip`.
+Name of the created extension package file. In this option, the values defined in `manifest.json` can be used by enclosing them with `{ }`. The default value is `{name}-{version}.zip`. The package file is always created in the artifacts directory specified by the global `--artifacts-dir` option.
```
rebloor: Rather than this change, I would suggest it makes more sense to add information to the introduction in line 126, like this:
```
By default, the package file is created in the relative path `./web-ext-artifacts.artifacts `. The location can be controlled with the [global](#global-options) `--artifacts-dir` option.
```

rebloor (changes requested):
@Zocker1999NET sorry for the delay.

src/content/documentation/develop/web-ext-command-reference.md
Zocker1999NET: you're right, thats the better location. I adapted the PR

rebloor:
Thanks @Zocker1999NET

PR #2316, opened 2026-04-25
Fix step 13 of Self-distribution

robbycbennett:
I clicked the link to the .xpi file as instructed, but it didn't actually download and save the file. Firefox skips that and tries to install it directly.

This was asked in https://support.mozilla.org/en-US/questions/961164.


rebloor:
Thanks @robbycbennett

PR #2319, opened 2026-04-30
Add numbering to each policy

bacharakis:
- Added a number to each policy and sub-sections to be easier to reference
- Replaces bullets with numbers

### DIFF
--- a/src/content/documentation/publish/add-on-policies.md
+++ b/src/content/documentation/publish/add-on-policies.md
@@ -266,9 +266,7 @@ If the primary function of the add-on does not work without transmitting persona
 
 #### 6.2.4. Implicit Consent for Self-Evident, Single-Use Extension
 
-Implicit consent applies only to add-ons hosted on addons.mozilla.org when all of the following conditions are satisfied. Otherwise the standard explicit consent rules apply.
-
-##### 6.2.4. Conditions
+Implicit consent applies only to add-ons hosted on addons.mozilla.org when all of these conditions are satisfied. Otherwise, the standard explicit consent rules apply.
 
 1. **Purpose-bounded and user-initiated** – Data may be transmitted only as a direct, immediate consequence of a single, deliberate user command (for example, a click or tap) on a clearly labelled control supplied by the browser or the add-on. Any passive, continuous, or background transmission requires explicit consent.
 
--- a/src/content/documentation/publish/add-on-policies.md
+++ b/src/content/documentation/publish/add-on-policies.md
@@ -15,9 +15,10 @@ contributors:
     wbamberg,
     kmaglione,
     dotproto,
+    bacharakis,
   ]
-last_updated_by: rebloor
-date: 2026-03-12
+last_updated_by: bacharakis
+date: 2026-04-30
 ---
 
 <!-- Page Hero Banner -->
@@ -43,7 +44,7 @@ When an add-on is given human review or otherwise assessed by Mozilla, these pol
 
 {% capture content_with_toc %}
 
-## No Surprises
+## 1. No Surprises
 
 Users should be able to easily discern the functionality of your add-on based on the listing, and should not be presented with unexpected user experiences after installing it. The listing should include an easy-to-read description of what the add-on does, and what information it transmits. Please consult our best practices guide for [creating an appealing listing](/documentation/develop/create-an-appealing-listing/).
 
@@ -55,15 +56,15 @@ Users should be able to easily discern the functionality of your add-on based on
 
 {% capture content %}
 
-### Unexpected features
+### 1.1. Unexpected features
 
 “Unexpected” features are those that are unrelated to the add-on’s primary function, and are not clearly indicated by the add-on name or description. This may include features that impact user privacy or security, make unexpected changes to web content, or change default settings like the new tab page, homepage, or search engine.
 
 Any “unexpected” feature(s) must adhere to all of the following requirements:
 
-* The add-on description must clearly state any changes made by these features.
-* The features must be “opt-in”, meaning the user has to take non-default action to enact the change. Changes prompted by Firefox after the add-on is installed do not require an additional opt-in. The permissions prompt shown when installing an add-on does not alleviate the need for an opt-in.
-* The opt-in interface must clearly state the name of the add-on requesting the change.
+1. The add-on description must clearly state any changes made by these features.
+2. The features must be “opt-in”, meaning the user has to take non-default action to enact the change. Changes prompted by Firefox after the add-on is installed do not require an additional opt-in. The permissions prompt shown when installing an add-on does not alleviate the need for an opt-in.
+3. The opt-in interface must clearly state the name of the add-on requesting the change.
 
 {% endcapture %}
 {% include modules/one-column.liquid,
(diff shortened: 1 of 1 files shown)


src/content/documentation/publish/add-on-policies.md
```diff
@@ -263,11 +264,11 @@ If the primary function of the add-on does not work without transmitting persona
 
 {% capture content %}
 
-#### Implicit Consent for Self-Evident, Single-Use Extension
+#### 6.2.4. Implicit Consent for Self-Evident, Single-Use Extension
 
 Implicit consent applies only to add-ons hosted on addons.mozilla.org when all of the following conditions are satisfied. Otherwise the standard explicit consent rules apply.
 
-##### Conditions
+##### 6.2.4. Conditions
```
rebloor: I'm not sure this warrants a numbered section and could be somewhat confusing, given it is followed by a numbered list that isn't in the same format. I think we can safely remove.
```suggestion
Implicit consent applies only to add-ons hosted on addons.mozilla.org when all of these conditions are satisfied. Otherwise, the standard explicit consent rules apply.
```

Otherwise, it needs to be numbered 6.2.4.1.

src/content/documentation/publish/add-on-policies.md
```diff
@@ -212,28 +213,28 @@ If the add-on uses native messaging, the Add-on Policies (including those relate
...
 
-#### If the add-on is only compatible with Firefox 140 or later and uses Firefox’s built-in data collection and transmission consent experience:
+#### 6.2.1. If the add-on is only compatible with Firefox 140 or later and uses Firefox’s built-in data collection and transmission consent experience:
 
 It must accurately state the data collection practices in the extension manifest, including when it does not collect data, in line with the [Firefox add-on data classification taxonomy](/documentation/develop/firefox-builtin-data-consent/#taxonomy).
 
-#### If the add-on is compatible with Firefox 139 and earlier or does not use Firefox’s built-in data collection and transmission consent experience:
+#### 6.2.2. If the add-on is compatible with Firefox 139 and earlier or does not use Firefox’s built-in data collection and transmission consent experience:
 
 The user must be provided with a clear way to control the add-on’s data transmission immediately after installation of the add-on. If data transmission starts or changes in an add-on update, or the consent and control is introduced in an update, it must be shown to all new and upgrading users immediately after the update.
```
rebloor, line 228: I'm not sure that these two items are policies, e.g., would you ever refer to them as "the If the add-on is compatible with Firefox 139 and earlier or does not use Firefox’s built-in data collection and transmission consent experience policy"? In the first instance, it seemed to be a statement that data collection requirements must be defined correctly in the built-in experience. While the second defines what the developer-created experience should cover. At the same time, it also seems to say you must use the built-in data collection experience for extensions that are compatible with only Firefox 140 or later. 
Therefore:
```suggestion
The user must be provided with a clear way to control the add-on’s data transmission, either through a [consent experience created by the add-on developer](/documentation/develop/best-practices-for-collecting-user-data-consents/) or by using [Firefox’s built-in data collection and transmission consent experience](/documentation/develop/firefox-builtin-data-consent/). Add-ons that are only compatible with Firefox 140 or later must use Firefox’s built-in data collection and transmission consent experience.

For add-ons that qualify for implicit consent under the [Implicit Consent for Self-Evident, Single-Use Extension](#implicit-consent-for-self-evident-single-use-extension) policy, installation counts as consent.

Add-ons installed in an enterprise environment can bypass asking for data collection consent when they are installed by enterprise policy. For more information, refer to the [enterprise documentation](/documentation/enterprise/enterprise-development/). If the add-on uses Firefox’s built-in data collection and transmission consent experience, then the browser will bypass this by default.

#### 6.2.1. Using Firefox’s built-in data collection and transmission consent experience. 

The add-on must accurately state the data collection practices in the extension manifest, including when it does not collect data, in line with the [Firefox add-on data classification taxonomy](/documentation/develop/firefox-builtin-data-consent/#taxonomy).

#### 6.2.2. Using a consent experience created by the add-on developer

The user must be provided with a clear way to control the add-on’s data transmission immediately after installation of the add-on. If data transmission starts or changes in an add-on update, or the consent and control is introduced in an update, it must be shown to all new and upgrading users immediately after the update.
```
Otherwise, I'd be inclined not to number these items and to leave the use of headings as typographical highlighting.

rebloor (commented):
@bacharakis I've avoided making text changes to the policies in the past, as I'm not sure whether for example they are reviewed and approved by legal. However, adding numbering to the headings throws up a couple of incongruities.

src/content/documentation/publish/add-on-policies.md
wagnerand-moz: > I'm not sure that these two items are policies, e.g., would you ever refer to them as "the If the add-on is compatible with Firefox 139 and earlier or does not use Firefox’s built-in data collection and transmission consent experience policy"? In the first instance, it seemed to be a statement that data collection requirements must be defined correctly in the built-in experience. While the second defines what the developer-created experience should cover.

That is accurate.

> At the same time, it also seems to say you must use the built-in data collection experience for extensions that are compatible with only Firefox 140 or later.

We are not requiring that at this time.

I understand what you are trying to resolve, but I would leave this for a later policy update and keep this to numbering only. These larger changes often require more coordination and approval.

src/content/documentation/publish/add-on-policies.md
wagnerand-moz: Agreed.

src/content/documentation/publish/add-on-policies.md
```diff
@@ -291,7 +292,7 @@ When all the above conditions are met, invoking the primary function is deemed i
 
 {% capture content %}
 
-#### Technical & User Interaction Data (opt-out)
+#### 6.2.5. Technical & User Interaction Data (opt-out)
```
wagnerand-moz: Up to (and including) here is actually all still 6.2.2 because these are all requirements for custom consent experiences only.

(Perhaps except this one sentence: "The add-on functionality must not be restricted if the user declines transmission of this data.", we can clean that up later.)

wagnerand-moz (changes requested):
Looks ok overall. The data collection numbering needs some changes.

src/content/documentation/publish/add-on-policies.md
rebloor: > I understand what you are trying to resolve, but I would leave this for a later policy update and keep this to numbering only. These larger changes often require more coordination and approval.

Which is why I preface my comments with

> I've avoided making text changes to the policies in the past, as I'm not sure whether for example they are reviewed and approved by legal. 

and didn't mark them as requested changes.

src/content/documentation/publish/add-on-policies.md
rebloor: However, don't these sections also describe what the policy considers each category of data to be, and is that also applicable to the built-in experience?

wagnerand-moz: Yes and no. These sections describe what the policies consider each category of data to be **for a custom consent**. For the built-in consent, we have streamlined the categories and criteria in https://extensionworkshop.com/documentation/develop/firefox-builtin-data-consent/.

At least that's what it meant to be. Do you see anything where that would be conflicting with what is on the page?

wagnerand-moz: @bacharakis you missed a couple of subheader numbers to remove.

rebloor: @wagnerand-moz the reasons why my reading suggests that the three sections (personal data, implicit consent, and technical data) apply to both consent experiences:

1. The opening paragraph mentions the two consent experiences and the implicit consent. It suggests that implicit consent is an alternative to both of the consent experiences. In addition, as the details about the implicit consent are presented between the content about the personal and technical data, further suggesting that these sections aren't specific to the developer-created experience.
2. The Implicit Consent for Self-Evident, Single-Use Extension section makes reference to the built-in experience data classification taxonomy.
3. The three sections are at the same heading level as the two choice options ("If the add-on is…")
4. The length of the "choice" statements doesn't make them seem like "heading", just emphasis on the text.

In addition to my earlier comments, e.g., changing the "choices" into more explicit headings, it may also be worth moving the definition information for the data types for a developer-created experience to the data classification section of the [best practices page](https://extensionworkshop.com/documentation/develop/best-practices-for-collecting-user-data-consents/).

I'd also be tempted to use the term "data classification" in both instances rather than "data classification" in one and "taxonomy" in the other.

bacharakis: @wagnerand-moz I have updated the numbers in the latest commit. I added the numbering without even reading the policy, by simply following the markdown hierarchy; that's why everything after: " If the add-on is compatible with Firefox 139 and earlier or does not use Firefox’s built-in data collection and transmission consent experience:" had a new number, because they had the same markdown hierarchy. 
I would like to keep the numbers for every paragraph if that's ok with you, so I can more easily refer to them to the developers.

wagnerand-moz: > it suggests that implicit consent is an alternative to both of the consent experiences.

It only applies to the custom consent as far as I know, it doesn't really make sense to apply to the built-in one as well, as that is integrated into the install doorhanger, and not an extra step, which is what we want to avoid for implicit cases. Something to keep in mind for fixing when we do a larger rewrite. At some point soon-ish (hopefully) the built-in consent will become mandatory anyway. 

> In addition, as the details about the implicit consent are presented between the content about the personal and technical data, further suggesting that these sections aren't specific to the developer-created experience.

We could consider simply swapping the order so that the personal and non-personal sections are next to each other.

> The Implicit Consent for Self-Evident, Single-Use Extension section makes reference to the built-in experience data classification taxonomy.

I agree that is not ideal. The implicit consent section was added very late in the process and we were trying to come up with something simple, so we used the more concrete taxonomy we had just build. The custom consent experience does not have such a concrete taxonomy. 

> The three sections are at the same heading level as the two choice options ("If the add-on is…")

That was for technical reasons. I tried to make the headings accurate when I made the changes for the policy update, but If I remember correctly, there were formatting issues caused by eleventy (or our styles we apply), and we deemed it not worth the effort and too risky to fix that, as it might cause other pages to change styling.

> The length of the "choice" statements doesn't make them seem like "heading", just emphasis on the text.

Initially they were indeed just emphasis, and that made things even more confusing, where it wasn't clear how sections apply to the different consents. We therefore changed them into headings (again not very visible due to the low header level and the subtle CSS style that is defined for those), but it seemed that was not enough.

> In addition to my earlier comments, e.g., changing the "choices" into more explicit headings, it may also be worth moving the definition information for the data types for a developer-created experience to the data classification section of the [best practices page](https://extensionworkshop.com/documentation/develop/best-practices-for-collecting-user-data-consents/).
> I'd also be tempted to use the term "data classification" in both instances rather than "data classification" in one and "taxonomy" in the other.

We can look into that when we do an actual update. Content-changes like those must go through several stakeholders internally first.

bacharakis: @wagnerand-moz we won't do any content changes for now, we will capture @rebloor feedback in a different doc and go from there.

wagnerand-moz (approved):

PR #2350, opened 2026-06-12
Fix create-an-appealing-listing permalink

### LINKED ISSUE #2349: NoSuchKey error for create-an-appealing-listing page
Something seems off with <https://extensionworkshop.com/documentation/publish/create-an-appealing-listing/>, which is linked to in the sidebar at <https://extensionworkshop.com/documentation/publish/>:

```
% curl https://extensionworkshop.com/documentation/publish/create-an-appealing-listing/
<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: extensionworkshop-prod-bucket/documentation/publish/create-an-appealing-listing/index.html</Details></Error>
```

I suppose this is the source for the page in question? <https://github.com/mozilla/extension-workshop/blob/master/src/content/documentation/publish/create-an-appealing-listing.md>

leoheitmannruiz:
Closes #2349

### DIFF
--- a/src/content/documentation/publish/create-an-appealing-listing.md
+++ b/src/content/documentation/publish/create-an-appealing-listing.md
@@ -2,7 +2,7 @@
 layout: sidebar
 title: Create an appealing listing
 description: Maximize your visibility with an appealing add-on listing. Learn how to write a compelling description and choose effective screenshots for addons.mozilla.org (AMO).
-permalink: /documentation/develop/create-an-appealing-listing/
+permalink: /documentation/publish/create-an-appealing-listing/
 topic: Publish
 tags: [add-ons, beginner, guide, publishing, webextension]
 contributors:


rebloor (approved):
@leoheitmannruiz Thanks for fixing this one

willdurand:
(I deployed this fix today, thanks both!)

rebloor:
Thanks @willdurand

willdurand:
I had to revert this change. It turns out that the permalink was fine, but the URL of the sidebar menu entry was incorrectly updated in #2197.

rebloor:
@willdurand, but the article should have a `/publish/` path - it's not a `/develop/` article.

willdurand:
> @willdurand, but the article should have a `/publish/` path - it's not a `/develop/` article.

Yeah, that was changed several years ago, but only the "Topic" was changed, not the URL, which is why we never set up any redirect. Now, I believe we should do that too, but we need a bunch of other changes, including some nginx config.

PR #2357, opened 2026-06-22
Use native fetch instead of node-fetch

willdurand:


diox (approved):
This seems harmless, but FWIW this should only affect hot reloading on local development, and _that_ doesn't seem to be working for me locally atm, even on `master` ...

PR #2358, opened 2026-06-22
Use npm instead of yarn

willdurand:
refs https://github.com/mozilla/addons/issues/16290


package.json
```diff
@@ -25,7 +25,7 @@
   "devDependencies": {
     "@11ty/eleventy": "3.1.6",
     "@11ty/eleventy-plugin-syntaxhighlight": "5.0.2",
-    "@eslint/js": "10.0.1",
+    "@eslint/js": "9.39.4",
```
diox, line 28: mmm? downgrading `@eslint/js` as part of this ?

willdurand: Yes. We have to because otherwise peer dependency resolution breaks. Sorry I forgot to mention it in the PR description.

diox (approved):

PR #2374, opened 2026-07-07
Add --enterprise to the list of web-ext CLI flags

willdurand:
refs https://github.com/mozilla/web-ext/issues/3771

### DIFF
--- a/src/content/documentation/develop/web-ext-command-reference.md
+++ b/src/content/documentation/develop/web-ext-command-reference.md
@@ -33,8 +33,8 @@ contributors:
     djbrown,
     Robot-Inventor
   ]
-last_updated_by: rebloor
-date: 2024-05-30
+last_updated_by: willdurand
+date: 2026-07-07
 ---
 
 <!-- Page Hero Banner -->
@@ -211,6 +211,19 @@ Environment variable: `$WEB_EXT_PRETTY=true`
 Declares that your extension will be self-hosted.&nbsp;This disables messages related to hosting on [addons.mozilla.org](https://addons.mozilla.org/).
 
 Environment variable: `$WEB_EXT_SELF_HOSTED=true`
+</section>
+
+<section id="enterprise">
+
+#### `--enterprise`
+
+Treats your extension as an enterprise extension when linting.
+
+::: note
+This option is new in web-ext 10.5.0.
+:::
+
+Environment variable: `$WEB_EXT_ENTERPRISE=true`
 </section>
 
 <section id="boring">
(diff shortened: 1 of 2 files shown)


src/content/documentation/develop/web-ext-command-reference.md
```diff
@@ -213,6 +213,19 @@ Declares that your extension will be self-hosted.&nbsp;This disables messages re
 Environment variable: `$WEB_EXT_SELF_HOSTED=true`
 </section>
 
+<section id="enterprise">
+
+#### `--enterprise`
+
+Treats your extension as an enterprise extension when linting.
```
rebloor: @willdurand, can we provide more information (or a link) that explains the effect of this, e.g., which linting is reduced or enhanced?

willdurand: I don't think we can unfortunately. The rest of the feature on AMO doesn't exist.

rebloor: I'm wondering why we are documenting something if we can't tell people what it does? Wouldn't it be better to wait until we can provide more information on the feature?

willdurand: We can always update the section once we have more information, but landing the doc with the feature that's already available seems like a good idea to me, still.

rebloor: @willdurand had a bit more of a dig here. I can see that the enterprise directive is included in the help. However, I still think that a feature for which we can't provide any information is not worth documenting. How can we track updates so we can add information as soon as it's available?
Possibly a more significant issue is that I noticed the version is now 10.5.0. But the documentation describes the current version as 8. I think we should update the version details and add `What's changed in Version x` for 9 and 10 to provide a summary of the significant changes.
Let me know if you'd like me to suggest changes to cover this.

willdurand: You're not wrong... It'd be interesting to have what's changed in v9 and v10, yes. Compared to v8, changes are lighter IIRC, and we release major versions mainly to bump the min supported Node version:

* https://github.com/mozilla/web-ext/releases/tag/9.0.0
* https://github.com/mozilla/web-ext/releases/tag/10.0.0

There's been a few other changes, though. If you could draft something, thta's be cool. Thanks!

> How can we track updates so we can add information as soon as it's available?

As for the above, I am not sure tbh. Maybe you're right and it's fine to leave this flag undocumented for now.

rebloor: @willdurand see https://github.com/mozilla/extension-workshop/pull/2384

rebloor (approved):

